PR #179, opened 2025-04-02
Better rate limiting across threads and better errors in failures

mahyareb:


datasets/bat/model_configs/gemini_2.0_pro_model.yaml
```diff
@@ -1,4 +1,4 @@
 generator: gcp_vertex_gemini
 vertex_model: gemini-2.0-pro-exp-02-05
```
totoleon, line 2: Is this still accessible from Vertex? It seems the 2.0 pro exp from AI Studio is no longer accessible, and it got replaced by the 2.5 pro exp.

totoleon (approved):

datasets/bat/model_configs/gemini_2.0_pro_model.yaml
mahyareb: Vertex seems to still work. We should add 2.5 here. Do you know if Vertex supports it?

PR #180, opened 2025-04-03
Support multi query statements and fix golden queries

jahnavi2k:


evalbench/databases/db.py
```diff
@@ -143,6 +144,16 @@ def batch_execute(self, commands: list[str]) -> None:
         """
         raise NotImplementedError("Subclasses must implement this method")
 
+    @abstractmethod
```
mahyareb, line 147: No need to make this an abstract method. You can just make the sub-classes define them as _ since it's only used internally by each class

evalbench/databases/mysql.py
```diff
@@ -87,6 +90,17 @@ def close_connections(self):
     #####################################################
     #####################################################
 
+    def execute_queries(self, connection: Connection, query: str) -> List:
+        result: List = []
+        for sub_query in sqlparse.split(query):
+            sub_query = sanitize_sql(sub_query)
```
mahyareb, line 96: Could you remove the use of sanitize_sql here and in other dbs?
We use Evalbench's execution for internal generations (google3) as well. Sanitize changes the expected SQL.

You could add this instead on the generator model here:
https://github.com/GoogleCloudPlatform/evalbench/blob/4f315c9cf19f444aa7c26a9647ddaee8c5e30798/evalbench/generators/models/gemini.py#L34

datasets/bat/prompts.json
```diff
@@ -1061,30 +1061,30 @@
...
+        "UPDATE tbl_attachments SET tags_file = json_insert(COALESCE(tags_file, '[]'), '$[#]', '\"data_security\"', '$[#]', '\"data_privacy\"'), category_file = json_insert(COALESCE(category_file, '[]'), '$[#]', '\"security\"', '$[#]', '\"privacy\"'), metadata = json_set(COALESCE(metadata, '{}'), '$.additional_info', '\"This attachment contains sensitive data related to security and privacy measures.\"') WHERE attachment_id = 1;UPDATE tbl_attachments SET tags_file = json_insert(COALESCE(tags_file, '[]'), '$[#]', '\"data_security\"', '$[#]', '\"data_privacy\"'), category_file = json_insert(COALESCE(category_file, '[]'), '$[#]', '\"security\"', '$[#]', '\"privacy\"'), metadata = json_set(COALESCE(metadata, '{}'), '$.additional_info', '\"This attachment contains sensitive data related to security and privacy measures.\"') WHERE attachment_id = 2;"
       ]
     },
     "eval_query": {
       "postgres": [
-        "Select * from tbl_attachments;"
+        "SELECT COUNT(*) FROM tbl_attachments WHERE attachment_id = 1;SELECT COUNT(*) FROM tbl_attachments WHERE attachment_id = 2;"
```
mahyareb, line 1078: Not sure if I understand what the intent of this is here. Could you give me some info?

mahyareb (commented):
Thank you!!

datasets/bat/prompts.json
jahnavi2k: Fixed them, sorry mb!

evalbench/databases/mysql.py
jahnavi2k: Done!

evalbench/databases/db.py
jahnavi2k: Makes sense, done!

evalbench/generators/models/gemini.py
```diff
@@ -32,13 +33,7 @@ def generate_internal(self, prompt):
...
             )
             if isinstance(response, GenerationResponse):
-                r = response.text
-                r = r.replace(
-                    "```sql", ""
-                )  # required for gemini_1.0_pro, gemini_2.0_flash, gemini_2.0_pro
-                r = r.replace(
-                    "```", ""
-                )  # required for gemini_1.0_pro, gemini_2.0_flash, gemini_2.0_pro
+                r = sanitize_sql(response.text)
```
mahyareb, line 36: Nice!! This looks much better now 😄

datasets/bat/prompts.json
```diff
@@ -1061,30 +1061,30 @@
...
+        "UPDATE tbl_attachments SET tags_file = COALESCE(tags_file, '[]'::jsonb) || '[\"data_security\", \"data_privacy\"]'::jsonb, category_file = COALESCE(category_file, '[]'::jsonb) || '[\"security\", \"privacy\"]'::jsonb, access_permissions = JSONB_SET( JSONB_SET(COALESCE(access_permissions, '{}'::jsonb), '{read_write}', 'true'::jsonb), '{read_only}', 'false'::jsonb), metadata = JSONB_SET(COALESCE(metadata, '{}'::jsonb), '{additional_info}', '\"This attachment contains sensitive data related to security and privacy measures.\"') WHERE attachment_id = 1;UPDATE tbl_attachments SET tags_file = COALESCE(tags_file, '[]'::jsonb) || '[\"data_security\", \"data_privacy\"]'::jsonb, category_file = COALESCE(category_file, '[]'::jsonb) || '[\"security\", \"privacy\"]'::jsonb, access_permissions = JSONB_SET( JSONB_SET(COALESCE(access_permissions, '{}'::jsonb), '{read_write}', 'true'::jsonb), '{read_only}', 'false'::jsonb), metadata = JSONB_SET(COALESCE(metadata, '{}'::jsonb), '{additional_info}', '\"This attachment contains sensitive data related to security and privacy measures.\"') WHERE attachment_id = 2;"
```
mahyareb, line 1064: can probably do these in a single query as ...... WHERE attachment_id IN (1,2); instead of multiple queries.

mahyareb (approved):
LGTM! Left one small nit comment. But feel free to raise a separate PR for that.

Thank you!

datasets/bat/prompts.json
jahnavi2k: Oh I meant to add different values for attachment ids 1 and 2. That's why I kept it separate. Let me know if we want to update both the ids with the same values.

mahyareb: Oh great! That sounds good to me!

mahyareb (approved):

PR #189, opened 2025-04-14
Add example colab for GCP CloudSQL Offerings (MySQL, Postgres, and SQL Server)

dvnshgupta:
GCP CloudSQL: https://colab.research.google.com/drive/1hlqB2C7rREIVfjdGIVxQh9fNgdsedS8T?resourcekey=0-0WWgMlsYCxzcAudbuE-WEw&usp=sharing


docs/examples/GCP_MySQL_example.ipynb
```diff
@@ -0,0 +1,410 @@
+{
+  "cells": [
+    {
+      "cell_type": "markdown",
+      "metadata": {
+        "id": "V5Bx7tBafUEP"
+      },
+      "source": [
+        "[![Open In Colab](https://colab.research.google.com/assets/colab-badge.svg)](https://colab.research.google.com/github/GoogleCloudPlatform/evalbench/blob/main/docs/examples/sqlite_example.ipynb)\n",
```
mahyareb, line 9: Could you update this link to link to itself?

docs/examples/GCP_MySQL_example.ipynb
```diff
@@ -0,0 +1,410 @@
...
+        "#### 6. Update the database config file"
+      ]
+    },
+    {
+      "cell_type": "markdown",
+      "metadata": {
+        "id": "dTH5n01ZE7P7"
+      },
+      "source": [
+        "**NOTE: Update Your DB name, DB path, username and password Below.**"
```
mahyareb, line 217: Somewhere between this step and starting, we should link to explain how to build a Cloud SQL instance. 

See the "Before you begin" section here and maybe replicate it?
https://colab.research.google.com/github/googleapis/langchain-google-cloud-sql-mssql-python/blob/main/docs/document_loader.ipynb#scrollTo=1Dpb9iXMF_cF

docs/examples/GCP_MySQL_example.ipynb
```diff
@@ -0,0 +1,410 @@
+{
```
mahyareb, line 1: Since all 3 examples are the exact same and just have a different db type, you can just merge them into one "GCP-CloudSQL-example" instead of 3 different files and let the user add their chosen DB type (with some instructions)

mahyareb (changes requested):
Thank you!

docs/examples/GCP_MySQL_example.ipynb
dvnshgupta: Thanks for pointing it out Mahyar!

docs/examples/GCP_MySQL_example.ipynb
dvnshgupta: That's a great suggestion Mahyar! I've replaced them with one single colab file.

PR #190, opened 2025-04-18
Adding support for BigQuery

jahnavi2k:
1. Adding support for BigQuery
2. Ensure that databases (without data) are created before executing DDL statements, so the schema is not empty when included in the model prompt.


datasets/bat/prompts.json
```diff
@@ -1,1456 +1,1657 @@
...
+          "ALTER TABLE tbl_tags ADD COLUMN access_level_enum ENUM('Public', 'Private', 'Restricted');"
+        ],
+        "sqlserver": [
+          "ALTER TABLE tbl_tags ADD access_level_enum VARCHAR(10) CONSTRAINT chk_access_level CHECK (access_level_enum IN ('Public', 'Private', 'Restricted'));"
+        ],
+        "sqlite": [
+          "ALTER TABLE tbl_tags ADD COLUMN access_level_enum TEXT CHECK (access_level_enum IN ('Public', 'Private', 'Restricted'));"
+        ],
+        "bigquery": [
+          "ALTER TABLE `{{dataset}}.tbl_tags1` ADD COLUMN access_level_enum STRING;"
```
mahyareb, line 1495: Might need to fix

datasets/bat/prompts.json
```diff
@@ -1,1456 +1,1657 @@
...
+        "sqlserver": [
+          "SELECT COUNT(DISTINCT u.user_id) AS num_bloggers FROM tbl_users u INNER JOIN tbl_posts p ON u.user_id = p.user_id WHERE p.created_at >= CAST('2024-01-01' AS DATE) AND p.created_at <= CAST('2024-12-31' AS DATE) AND JSON_VALUE(u.social_media_links, '$.twitter') IS NOT NULL;"
+        ],
+        "sqlite": [
+          "SELECT COUNT(DISTINCT u.user_id) AS num_bloggers FROM tbl_users u INNER JOIN tbl_posts p ON u.user_id = p.user_id WHERE p.created_at BETWEEN DATE('2024-01-01') AND DATE('2024-12-31') AND json_extract(u.social_media_links, '$.twitter') IS NOT NULL;"
+        ],
+        "bigquery": [
+          "SELECT COUNT(DISTINCT u.user_id) AS num_bloggers FROM {{dataset}}.tbl_users u INNER JOIN {{dataset}}.tbl_posts p ON u.user_id = p.user_id WHERE p.created_at BETWEEN DATE('2024-01-01') AND DATE('2024-12-31') AND JSON_EXTRACT_SCALAR(u.social_media_links, '$.twitter') IS NOT NULL;SELECT u.user_id, p.post_id, COUNT(c.comment_id) AS comment_count FROM `{{dataset}}.tbl_users` AS u INNER JOIN `{{dataset}}.tbl_posts` AS p ON u.user_id = p.user_id LEFT JOIN `{{dataset}}.tbl_comments` AS c ON p.post_id = c.post_id GROUP BY u.user_id, p.post_id;"
```
mahyareb, line 28: Seems like we have two statements here? Delete the second?

evalbench/databases/bigquery.py
```diff
@@ -0,0 +1,249 @@
...
+        return True
+
+    def insert_data(self, data: dict[str, List[str]]):
+        if not data:
+            return
+        insertion_statements = []
+        for table_name in data:
+            for row in data[table_name]:
+                formatted_values = []
+                for value in row:
```
mahyareb, line 212: Could you document this one? Do we have TRUE / FALSE values as '1' and '2' today?

Could you maybe check to make sure the column (str in dict) is a boolean type?

mahyareb (changes requested):
Overall looks good! Thank you very much Jahnavi. Left a few minor comments.

One refactoring request:
Instead of replacing the {{dataset}} in multiple places such as config.py and sqlexecwork.py, can you just do it in the bigquery.py's _execute function?

datasets/bat/prompts.json
jahnavi2k: Done!

datasets/bat/prompts.json
jahnavi2k: There's no enum in bigquery. Not sure if this is enough.

datasets/bat/prompts.json
jahnavi2k: Done!

datasets/bat/prompts.json
jahnavi2k: Done!

datasets/bat/prompts.json
```diff
@@ -1,1456 +1,1657 @@
...
+        ],
+        "sqlserver": [
+          "SELECT SUM(likes_count) FROM tbl_comments WHERE comment_id IN (SELECT reported_comment_id FROM tbl_reports WHERE is_resolved = 0 AND user_id IN (SELECT following_id FROM tbl_followers GROUP BY following_id HAVING COUNT(follower_id) > 4) AND post_id IN (SELECT post_id FROM tbl_posts_tags INNER JOIN tbl_tags ON tbl_posts_tags.tag_id = tbl_tags.tag_id WHERE name = 'Programming'))"
+        ],
+        "sqlite": [
+          "SELECT SUM(likes_count) FROM tbl_comments WHERE comment_id IN (SELECT reported_comment_id FROM tbl_reports WHERE is_resolved = 0 AND user_id IN (SELECT following_id FROM tbl_followers GROUP BY following_id HAVING COUNT(follower_id) > 4) AND post_id IN (SELECT post_id FROM tbl_posts_tags INNER JOIN tbl_tags ON tbl_posts_tags.tag_id = tbl_tags.tag_id WHERE name = 'Programming'));"
+        ],
+        "bigquery": [
+          "SELECT u.user_id, COUNT(f.follower_id) AS follower_count FROM `{{dataset}}.tbl_users` AS u LEFT JOIN `{{dataset}}.tbl_followers` AS f ON u.user_id = f.following_id GROUP BY u.user_id;"
```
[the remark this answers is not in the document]
jahnavi2k: Done!

evalbench/databases/bigquery.py
jahnavi2k: > Could you maybe check to make sure the column (str in dict) is a boolean type?
    If I am not wrong, the string in the dict is the table name and not the column? 
    
> Could you document this one? Do we have TRUE / FALSE values as '1' and '2' today?
    a. As of now we have values like true / false and ‘1’/’0’ for booleans, which cannot be identified by python as 
    booleans directly. Hence we are explicitly checking it. true/false get identified as JSON if not checked explicitly.
    b. As for JSONs, we need to unquote it and then check if it a json value.
    c. It recognizes datetime data correctly on its own, hence no separate check for it.

evalbench/util/config.py
```diff
@@ -41,13 +41,15 @@ def load_db_data_from_csvs(data_directory: str):
     return tables
 
 
-def load_setup_scripts(setup_scripts_directory_path: str):
+def load_setup_scripts(setup_scripts_directory_path: str, db_name: str):
```
mahyareb, line 44: I think you can just remove db_name here, since it's not used.
And just revert the file changes.

evalbench/evaluator/db_manager.py
```diff
@@ -61,36 +62,45 @@ def _prepare_db_queue_for_dml(core_db: DB, db_name, db_config, setup_config, num
...
     tmp_db.set_setup_instructions(setup_scripts, None)
     return tmp_db
 
 
-def _get_setup_values(setup_config, db_name: str, db_type: str):
+def _get_setup_values(setup_config, db_name: str, db_type: str, tmp_db: Optional[str] = None):
     try:
         setup_scripts = load_setup_scripts(
-            setup_config["setup_directory"] + "/" + db_name + "/" + db_type
+            setup_config["setup_directory"] + "/" + db_name + "/" + db_type, tmp_db if tmp_db is not None else db_name
```
mahyareb, line 103: can remove ", tmp_db if tmp_db is not None else db_name"

mahyareb (approved):
LGTM! Thank you!

evalbench/databases/bigquery.py
mahyareb: Talked offline, sounds like we can just re-use the resetup values to check if column is boolean here. Thanks Jahnavi!

jahnavi2k: Done!

evalbench/evaluator/db_manager.py
```diff
@@ -61,33 +62,42 @@ def _prepare_db_queue_for_dml(core_db: DB, db_name, db_config, setup_config, num
...
+            _create_ddl_tmp_db, db_name=db_name, db_config=db_config, setup_config=setup_config
         )
         results = executor.map(create_ddl_tmp_db_p, tmp_dbs)
         for tmp_db in results:
             db_queue.put(tmp_db)
     return db_queue
 
 
-def _create_ddl_tmp_db(db_name, db_config, setup_scripts):
+def _create_ddl_tmp_db(tmp_db, db_name, db_config, setup_config):
```
mahyareb, line 89: A few minor things I noticed, this has tmp_db but it's not being set anywhere (line 81)
Should we just use db_name?

evalbench/evaluator/db_manager.py
```diff
@@ -61,33 +62,42 @@ def _prepare_db_queue_for_dml(core_db: DB, db_name, db_config, setup_config, num
...
     tmp_ddl_db_config = deepcopy(db_config)
     tmp_ddl_db_config["is_tmp_db"] = True
-    tmp_db = get_database(tmp_ddl_db_config, db_name)
+    tmp_db = get_database(tmp_ddl_db_config, tmp_db)
     tmp_db.set_setup_instructions(setup_scripts, None)
     return tmp_db
 
 
-def _get_setup_values(setup_config, db_name: str, db_type: str):
+def _get_setup_values(setup_config, db_name: str, db_type: str, tmp_db: Optional[str] = None):
```
mahyareb, line 100: tmp_db here seems to be just thrown away? Should we just remove this parameter?

mahyareb (dismissed):
Thanks Jahnavi. Looks great. Left a couple of small questions on db_manager that I didn't understand well.

evalbench/evaluator/db_manager.py
jahnavi2k: We actually don't need db_name to be passed since we are no longer replacing the {{dataset}} here. We do need to pass the tmp_db though which is getting passed through [this line].(https://github.com/GoogleCloudPlatform/evalbench/blob/4dc06e96ac17971839c2cbc67d7bb018c0ba66bf/evalbench/evaluator/db_manager.py#L83). We can rename tmp_db back to db_name, but I feel it is clearer this way. 

Please let me know if I misunderstood your question!

mahyareb (approved):
Thank you!

mahyareb (approved):

PR #210, opened 2025-09-25
Interact

IsmailMehdi:
This change adds support for multi-turn eval. Specifically BIRD-INTERACT mode.
it adds:

- Specific orchestrator
- Specific evaluator flow
- Specific prompts
- Virtual user to disambiguate.


evalbench/evaluator/__init__.py
```diff
@@ -0,0 +1,15 @@
+from evaluator.orchestrator import Orchestrator
+from evaluator.oneshotorchestrator import OneShotOrchestrator
+from evaluator.interactorchestrator import InteractOrchestrator
+import logging
+
+
+def get_orchestrator(config, db_configs, setup_config, report_progress=False):
+    orchestrator_type = config.get("orchestrator", "oneshot")
+    logging.info(f"Orchestrator Type: {orchestrator_type}")
+    if orchestrator_type == "oneshot":
```
mahyareb, line 10: Should we make this default?

mahyareb (approved):

evalbench/evaluator/__init__.py
IsmailMehdi: it will since config.get("orchestrator", "oneshot") will return oneshot if not specified.

PR #307, opened 2026-04-06
fix: make PYTHONPATH robust across different working directories

aseering:
Update various launcher scripts to consistently set PYTHONPATH to point to the evalbench directory, relative to the path of that script (so the script still works correctly if called with a current working directory of some other directory).

Update protobufs to treat evalbench as the package root, so we don't additionally have to add their directory to PYTHONPATH in code that imports compiled-protobuf files.

Update a couple places that had accidentally ended up relying on the project root being in PYTHONPATH (so imported from `evalbench.*`).

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -197,10 +197,10 @@ pod-shell-test:
 
 proto:
 	@python -m grpc_tools.protoc \
-		--proto_path=evalbench/evalproto \
-		--python_out=evalbench/evalproto \
-		--pyi_out=evalbench/evalproto \
-		--grpc_python_out=evalbench/evalproto \
+		--proto_path=evalbench \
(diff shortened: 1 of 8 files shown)


IsmailMehdi:
i need some significant testing for this, it'll take time.

aseering:
No rush, from my perspective at least half of it is just a clean-up of my own code (and the other more-exciting half isn't blocking anything that I know of).

IsmailMehdi (approved):

saurabh-net:
Just curious, is this more relevant now that we have PRs which change the working directory?

https://github.com/GoogleCloudPlatform/evalbench/pull/349
https://github.com/GoogleCloudPlatform/evalbench/pull/355

PR #336, opened 2026-04-20
feat: add support for custom host configuration and insecure gRPC channels via environment variables

omkargaikwad23:
# Add environment variable overrides for gRPC server binding and insecure credentials

## Context & Problem
In containerized CI environments (such as Google Cloud Build), network configurations often do not route IPv6 loopback (`[::]`) or require binding to all available IPv4 interfaces (`0.0.0.0`) for reliable inter-process communication. 

Previously, `eval_server.py` and `eval_client.py` strictly hardcoded the gRPC server interface to `[::]` and mandated secure ALTS channel credentials for local endpoints. This required inline `sed` patching in downstream repository CI pipelines to execute integration tests successfully.

## Solution
This PR introduces native environment variable overrides to support flexible network bindings without requiring code patching downstream:

1. **Configurable Server Host:**  
   - Added support for `EVALBENCH_HOST` in `eval_server.py` (defaults to `[::]`). Setting this to `0.0.0.0` binds the server to all network interfaces.
   - Added an explicit check to raise a `RuntimeError` if port binding fails, enabling fail-fast behavior in CI pipelines.

2. **Insecure Local Credentials Override:**  
   - Added support for `EVALBENCH_INSECURE=true` in `eval_client.py` to bypass ALTS channel credentials when running locally or in offline test environments.

### DIFF
--- a/evalbench/client/eval_client.py
+++ b/evalbench/client/eval_client.py
@@ -1,4 +1,5 @@
 import asyncio
+import os
 from aiologger import Logger
 import grpc
 from evalproto import eval_request_pb2, eval_connect_pb2, eval_config_pb2
@@ -28,7 +29,6 @@ class EvalbenchClient:
     def __init__(self, endpoint: str):
         self.endpoint = endpoint
         if self.endpoint == "local":
-            import os
             host = os.getenv("EVALBENCH_HOST", "localhost")
             address = f"{host}:50051"
             if os.getenv("EVALBENCH_INSECURE", "").lower() == "true":
--- a/evalbench/client/eval_client.py
+++ b/evalbench/client/eval_client.py
@@ -28,9 +28,14 @@ class EvalbenchClient:
     def __init__(self, endpoint: str):
         self.endpoint = endpoint
         if self.endpoint == "local":
-            address = "localhost:50051"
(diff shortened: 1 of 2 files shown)


evalbench/client/eval_client.py
```diff
@@ -28,9 +29,13 @@ class EvalbenchClient:
     def __init__(self, endpoint: str):
         self.endpoint = endpoint
         if self.endpoint == "local":
-            address = "localhost:50051"
-            channel_creds = grpc.alts_channel_credentials()
-            self.channel = grpc.aio.secure_channel(address, channel_creds)
+            host = os.getenv("EVALBENCH_HOST", "localhost")
```
omkargaikwad23, line 32: @IsmailMehdi Should we also add flexibility to change the PORT number (e.g., via EVALBENCH_PORT) while keeping the default as 50051? This would complete the external configuration for the server binding.

IsmailMehdi (dismissed):

evalbench/client/eval_client.py
omkargaikwad23: Added this support.

IsmailMehdi (approved):

PR #338, opened 2026-04-21
fix(geminicli): resolve extension names from local manifests to apply settings

omkargaikwad23:
#### Problem
When installing a local extension (e.g., using a path like `/workspace` or `.`), `evalbench` previously derived the extension ID solely by applying a regex to the path string. This works for standard Git URLs, but results in an incorrect identifier for local deployments (extracting `"workspace"` instead of the true name `"cloud-sql-postgresql"`). 

As a result, the post-install configuration logic fails to map critical setup variables and database credentials to the agent. 

#### Solution
Updated the extension installation pipeline in `gemini_cli.py` to:
1. Look for a local `gemini-extension.json` manifest at the target path.
2. Parse the explicit `"name"` attribute from the manifest.
3. Fall back to standard string parsing for external Git references. 

This guarantees that custom credentials and target variables execute successfully on local evaluations.

### DIFF
--- a/evalbench/generators/models/gemini_cli.py
+++ b/evalbench/generators/models/gemini_cli.py
@@ -651,25 +651,37 @@ class GeminiCliGenerator(QueryGenerator):
                 else:
                     logging.info(f"Successfully installed extension: {ext}")
 
-                ext_name_match = re.search(r"([^/]+?)(?:\.git)?$", ext)
-                if ext_name_match:
-                    search_name = ext_name_match.group(1)
-                    if os.path.exists(self.extensions_dir):
-                        for item in os.listdir(self.extensions_dir):
-                            if search_name in item:
-                                ext_dir = os.path.join(self.extensions_dir, item)
-                                self._patch_manifest_sensitive(ext_dir)
-                                # For gemini-cli >= 0.36.0, --skip-settings was used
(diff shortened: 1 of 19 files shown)


evalbench/generators/models/gemini_cli.py
```diff
@@ -651,25 +651,37 @@ def _install_extensions(self, extensions: dict | list):
...
-                                    settings_values = (
-                                        extensions[ext].get("settings", {})
-                                        if extensions[ext]
-                                        else {}
-                                    )
-                                    self._configure_extension_settings(
-                                        item, ext_dir, settings_values, install_env
-                                    )
+                search_name = None
+                manifest_path = os.path.join(ext, "gemini-extension.json")
```
prernakakkar-google, line 655: This might fail for git paths

IsmailMehdi (approved):
lgtm

evalbench/generators/models/gemini_cli.py
omkargaikwad23: We have fallback mechanism to search in GitHub URL if not found in gemini-extension.json 
But, I think I should prioritize extracting the extension name via regex from the GitHub URL, with the local gemini-extension.json manifest serving as the fallback. This ensures we support both local and remote path configurations.

PR #344, opened 2026-04-25
Feature/python scorer

saurabh-net:
# PR Description: Add General PythonScorer using `uv run`

## Goal
This PR introduces a general-purpose `PythonScorer` that allows users to author custom evaluation logic in standalone Python scripts. This removes the need to check every custom scorer into the core EvalBench repository, facilitating rapid experimentation.

### evalbench
*   **[NEW] `evalbench/scorers/pythonscorer.py`**:
    *   Implements `PythonScorer` class inheriting from `Comparator`.
    *   Delegates evaluation to an external script specified by `script_path` in the configuration.
    *   Uses **`uv run`** to execute the script as a subprocess.
    *   **Data Contract**: Passes the full evaluation context (as received by `Comparator.compare`) to the script via **standard input** as a JSON string. Expects the script to output a JSON object to **standard output** containing `score` (float) and `reason` (string).
    *   **Isolation**: By using `uv run`, it supports PEP 723 (inline script metadata), allowing user scripts to declare their own dependencies (e.g., `pandas`, `rich`) which `uv` will automatically install in an isolated environment at runtime.
    *   **Fallback handling**: If `uv` is not found in the environment, it fails gracefully with a helpful message: `'uv' command not found in PATH`.
*   **`evalbench/scorers/score.py`**:
    *   Registered `python_scorer` in the `compare` function.

### Automated Tests
*   Created `evalbench/test/pythonscorer_test.py` covering:
    *   Successful execution and JSON parsing (mocked subprocess).
    *   Handling of script failures (non-zero exit code).
    *   Handling of invalid JSON output from the script.
    *   Handling of missing `uv` command.
*   All tests passed successfully.

### Manual Verification
*   **Verified without `uv`**: Confirmed that the fallback correctly triggers and reports a clean failure message (`'uv' command not found in PATH`) when `uv` was missing from the environment.
*   **Verified with `uv`**: After installing `uv` in the environment, confirmed that `uv run` successfully handled inline script dependencies (PEP 723) and executed the custom script, returning the expected score and reason.


IsmailMehdi (dismissed):

saurabh-net:
Tested this manually as well. LGTM.

IsmailMehdi (approved):

PR #356, opened 2026-04-29
feat: handle recursive dependency cleanup in Spanner table drop

aseering:
Spanner's support for DROP ... CASCADE is quite limited right now.  If for example a DDL Eval creates a view and then doesn't drop it (and I've found an instance of this in one of our datasets), EvalBench is then unable to drop any tables queried by that view.  This CL adds more-proactive dropping.


patch_spanner_lint.py
IsmailMehdi, line 1: i don't think we need this file

IsmailMehdi (dismissed):

IsmailMehdi:
please fix style issues

IsmailMehdi (approved):

PR #358, opened 2026-04-29
feat: use parameterized queries for data insertion in MySQL, Postgres, and SQLite

aseering:
This approach means that we don't have to worry about how to properly quote data, now that datasets are no longer pre-wrapped in single quotes for us.  It also has a modest performance benefit.  (It would have a big performance benefit for Spanner, but Spanner's implementation already uses an entirely separate API for data ingest.)

If we want to, we could move to prepared INSERT everywhere including for Spanner.  Might be nice for consistency.  Spanner does not have good throughput with non-prepared INSERTs, but prepared INSERTs work quite well.  That would be a larger project, though, and I think not urgent.


evalbench/databases/sqlite.py
```diff
@@ -249,22 +243,35 @@ def drop_all_tables(self):
...
+                            p[f"v{i}"] = self._clean_insert_value(val)
+                        params.append(p)
+
+                    connection.execute(stmt, params)
+        except Exception as error:
             raise RuntimeError(f"Could not insert data into database: {error}")
 
-    #####################################################
+    def _format_boolean_value(self, val: str) -> Any:
+        return "true" if val == "true" else "False"
```
IsmailMehdi: any reason the values capitalization is not consistent?

aseering: 'Cause silly Python reflexes.  Fixed.

IsmailMehdi (approved):

PR #370, opened 2026-05-05
Add support for bidirectional streaming within eval service.

erinlimbogoogle:


evalbench/evalproto/eval_request.proto
```diff
@@ -16,6 +16,7 @@ message DialectBasedSQLStatements {
 
 message EvalInputRequest {
   int64 id = 1;
+  string conversation_id = 18;
```
IsmailMehdi: should be sorted by id

evalbench/evalproto/eval_request.proto
```diff
@@ -29,6 +30,7 @@ message EvalInputRequest {
   string sql_generator_error = 12;
   float sql_generator_time = 13;
   string generated_sql = 14;
+  string generated_nl_response = 19;
```
IsmailMehdi: same

evalbench/eval_service.py
```diff
@@ -241,6 +245,169 @@ async def Eval(
...
+                first_item = dataset[0]
+                conv_id = str(getattr(first_item, "id", ""))
+                if conv_id:
+                    final_request.conversation_id = conv_id
+            logging.info(f"Yielding final summary payload: {final_request}")
+            yield final_request
+
+        finally:
+            # Clean up the global registry to prevent memory leaks.
+            PROXY_QUEUES.pop(session_id, None)
```
IsmailMehdi, line 438: If a client connection drops forcefully or hangs, and the  Interact  RPC does not hit its  finally  block cleanup (or if the thread hangs indefinitely
  blocked on  inbound_response = thread_inbox.get(block=True, timeout=300.0) ), then that entry in  PROXY_QUEUES  will leak and remain in memory even
  after the background reaper deletes the session from Disk/SessionManager!

we should clear it when the session gets deleted.

erinlimbogoogle: Ah thanks for the catch. I added a context.add_done_callback to cleanup on forceful disconnects, and added a heartbeat check in the polling loop to terminate the stream and clear the queue if the SessionManager reaper deletes the session.

evalbench/evalproto/eval_request.proto
erinlimbogoogle: Done.

evalbench/evalproto/eval_request.proto
erinlimbogoogle: Done.

IsmailMehdi (approved):

PR #396, opened 2026-05-18
add support for skills in codex

prernakakkar-google:


datasets/codex-skills/codex_skills_run_config.yaml
```diff
@@ -0,0 +1,24 @@
...
+
+runners:
+  agent_runners: 1
+
+scorers:
+  skills_trajectory:
+    enforce_order: false
+  skills_best_practices:
+    model_config: datasets/model_configs/gemini_3.1_pro_model.yaml
+    skills_dir: ".venv/fake_home_codex/.codex/plugins/marketplaces/cloud-sql-postgresql/skills"
```
IsmailMehdi: why is this here ? it should be inferred.

evalbench/scorers/skillsbestpractices.py
```diff
@@ -37,20 +37,25 @@ def __init__(self, config: dict, global_models):
...
-            fake_home_skills = os.path.join(
-                ".venv", "fake_home_claude", ".claude", "skills")
-            if os.path.isdir(fake_home_skills):
-                self.skills_dir = os.path.abspath(fake_home_skills)
-                logging.info(
-                    f"Using fake_home skills directory: {self.skills_dir}")
+            fallback_skill_dirs = (
+                os.path.join(".venv", "fake_home_claude", ".claude", "skills"),
+                os.path.join(".venv", "fake_home_codex", ".codex", "skills"),
+                os.path.join(".venv", "fake_home_gemini", ".gemini", "skills"),
```
IsmailMehdi: is this the right name, i see fake_home.

IsmailMehdi (commented):
Reviewing the Codex skills implementation. Mostly looks great, but found a few minor issues.

evalbench/scorers/skillsbestpractices.py
```diff
@@ -37,20 +37,25 @@ def __init__(self, config: dict, global_models):
 
         # Resolve where skill SKILL.md files live. Prefer an explicit
         # `skills_dir` in the scorer config; otherwise fall back to the
```
IsmailMehdi, line 39: This path uses `fake_home_gemini`, but the Gemini generator (`gemini_cli.py`) defines its fake home as `fake_home` (without the `_gemini` suffix). It should be `fake_home` here as well.

Alternatively, we can resolve `skills_dir` dynamically using the `fake_home` path passed in the `generated_eval_result` context, which would also solve the issue when running with `session_id` prefix (under `/tmp_sessions/`).

datasets/codex-skills/codex_skills_run_config.yaml
```diff
@@ -0,0 +1,24 @@
...
+orchestrator: agent
+model_config: datasets/codex-skills/codex_skills_model.yaml
+simulated_user_model_config: datasets/model_configs/gemini_3.1_pro_model.yaml
+
+runners:
+  agent_runners: 1
+
+scorers:
+  skills_trajectory:
+    enforce_order: false
```
IsmailMehdi, line 13: Is `marketplaces` correct here? The generator (`codex_cli.py`) clones the repo to `plugins/cloud-sql-postgresql` (without `marketplaces`). Setting this explicitly might cause the scorer to fail to find the skills.

If you omit `skills_dir` entirely, it will automatically fall back to the correct path `.venv/fake_home_codex/.codex/skills` where the skills are copied during setup.

IsmailMehdi (commented):
Thanks for addressing the earlier round — the dynamic `context["fake_home"]` resolution in `skillsbestpractices.py` is a nice cleanup and the `fake_home`-vs-`fake_home_gemini` fix is in. Dropping the hardcoded `marketplaces/cloud-sql-postgresql` `skills_dir` from the run config is also the right call.

A few things to look at on the latest commit:

**Blocker**
- `datasets/bird-interact/bird-interact-lite` submodule is being deleted in this PR (`387a6d1` "Fix: convert bird-interact-lite from broken submodule to regular directory" — but the directory isn't re-added in the diff). That's unrelated to Codex skills and removes the only reference to the submodule. Either split it into its own PR with the directory content re-added, or drop it from this branch via `git rebase`.

**Questions / verify**
- `codex_cli.py:353` — `_register_codex_plugin` writes the marketplace JSON to `<fake_home>/.agents/plugins/marketplace.json`. Codex CLI's plugin docs I've seen put marketplaces under `~/.codex/plugins/`. Is the `.agents/` path something a recent Codex version reads from, or a leftover from an experiment? If it's intentional, please link the Codex source or doc — otherwise the cloned repo will sit in `.codex/plugins/` unregistered and `expected_skills` matching will rely entirely on the `_install_skills_from_source` copy.
- `codex_skills_model.yaml:3` — `model: "gpt-5.5"`. Is that the intended model? Codex models I've seen are `gpt-5`, `gpt-5-codex`. If it's a typo it'll fail loudly at first run, but worth confirming before merge.

**Code quality**
- `codex_cli.py:1066` — `add_skill` falls open when `installed_skills` is empty: `if name and (not installed_skills or name in installed_skills) ...`. If `_get_installed_skills()` returns an empty set for any reason (race, permission, wrong path), every tool name gets reported as an activated skill, which then inflates the trajectory + best-practices scorers. I'd flip this to fail closed — return `[]` when `installed_skills` is empty.

**Nit**
- `codex_skills_model.yaml:12-15` — `CLOUD_SQL_POSTGRES_INSTANCE: <instance>` etc. are unquoted literal placeholders. As written, the eval will run with the strings `<instance>`, `<database>`, etc. in those env vars. Either drop them with a `# Set these per scenario` comment, or quote them as `"<instance>"` and add a note that users must substitute.

**Carry-over (still open from prior round)**
- `codex_skills_run_config.yaml:15` (`skills_best_practices.model_config`) — same comment as before, the per-scorer `model_config` looks like it should be inferable from a default. Not blocking if this is the project-wide pattern (other scorers in the same file repeat it), but worth a short reply to close the thread.

LGTM otherwise on the Codex-side wiring once the submodule deletion is removed.

IsmailMehdi:
Walking back my earlier "blocker" comment about the `bird-interact-lite` submodule deletion — apologies, I was wrong. Digging into it: `datasets/bird-interact/download_dataset.sh:21` clones `https://huggingface.co/datasets/birdsql/bird-interact-lite` into that path at runtime, so the directory was never meant to be tracked. What got committed (back in #212) was a stray gitlink — `git add` on the freshly-cloned directory recorded the embedded `.git/` as a submodule pointer, with no `.gitmodules` entry to back it. That's why `git submodule update --init` fails with `No url found for submodule path 'datasets/bird-interact/bird-interact-lite' in .gitmodules`. Your deletion is the right fix, not a blocker.

I pushed `c50dfcb` to add `datasets/bird-interact/bird-interact-lite/` to `.gitignore` so re-running `download_dataset.sh` can't recreate the trap. Feel free to revert if you'd rather handle it separately.

Other review comments (marketplace path, `extract_skills` fail-open, placeholder env vars, model name) still stand.

prernakakkar-google:
> Thanks for addressing the earlier round — the dynamic `context["fake_home"]` resolution in `skillsbestpractices.py` is a nice cleanup and the `fake_home`-vs-`fake_home_gemini` fix is in. Dropping the hardcoded `marketplaces/cloud-sql-postgresql` `skills_dir` from the run config is also the right call.
> 
> A few things to look at on the latest commit:
> 
> **Blocker**
> 
> * `datasets/bird-interact/bird-interact-lite` submodule is being deleted in this PR (`387a6d1` "Fix: convert bird-interact-lite from broken submodule to regular directory" — but the directory isn't re-added in the diff). That's unrelated to Codex skills and removes the only reference to the submodule. Either split it into its own PR with the directory content re-added, or drop it from this branch via `git rebase`.
> 
> **Questions / verify**
> 
> * `codex_cli.py:353` — `_register_codex_plugin` writes the marketplace JSON to `<fake_home>/.agents/plugins/marketplace.json`. Codex CLI's plugin docs I've seen put marketplaces under `~/.codex/plugins/`. Is the `.agents/` path something a recent Codex version reads from, or a leftover from an experiment? If it's intentional, please link the Codex source or doc — otherwise the cloned repo will sit in `.codex/plugins/` unregistered and `expected_skills` matching will rely entirely on the `_install_skills_from_source` copy.
> * `codex_skills_model.yaml:3` — `model: "gpt-5.5"`. Is that the intended model? Codex models I've seen are `gpt-5`, `gpt-5-codex`. If it's a typo it'll fail loudly at first run, but worth confirming before merge.
> 
> **Code quality**
> 
> * `codex_cli.py:1066` — `add_skill` falls open when `installed_skills` is empty: `if name and (not installed_skills or name in installed_skills) ...`. If `_get_installed_skills()` returns an empty set for any reason (race, permission, wrong path), every tool name gets reported as an activated skill, which then inflates the trajectory + best-practices scorers. I'd flip this to fail closed — return `[]` when `installed_skills` is empty.
> 
> **Nit**
> 
> * `codex_skills_model.yaml:12-15` — `CLOUD_SQL_POSTGRES_INSTANCE: <instance>` etc. are unquoted literal placeholders. As written, the eval will run with the strings `<instance>`, `<database>`, etc. in those env vars. Either drop them with a `# Set these per scenario` comment, or quote them as `"<instance>"` and add a note that users must substitute.
> 
> **Carry-over (still open from prior round)**
> 
> * `codex_skills_run_config.yaml:15` (`skills_best_practices.model_config`) — same comment as before, the per-scorer `model_config` looks like it should be inferable from a default. Not blocking if this is the project-wide pattern (other scorers in the same file repeat it), but worth a short reply to close the thread.
> 
> LGTM otherwise on the Codex-side wiring once the submodule deletion is removed.

https://openai.com/index/introducing-gpt-5-5/ - GPT 5.5 is the latest model and was using it.
Fixes applied:
.agents path → .codex (codex_cli.py:69,352, docs/codex_cli_agent_testing.md:382) — marketplace.json now lives next to the cloned repo at <fake_home>/.codex/plugins/marketplace.json. Removed the unused codex_agents_dir attribute. Updated the doc to match.

add_skill fails closed (codex_cli.py:1063-1070) — extract_skills now returns [] early when installed_skills is empty instead of accepting every tool name.

YAML placeholders quoted + commented (codex_skills_model.yaml:9-16) — added a "substitute before running" comment; quoted <instance>/<database>/<user>/<password> so they're clearly placeholder strings.

fir models will do a separate change as its done throughout and would requires us to change everywhere including CI.

IsmailMehdi (approved):

PR #402, opened 2026-05-22
feat(scorers): filter native tools in trajectory_matcher with opt-out flag

omkargaikwad23:
## Summary
  -  Add looks_like_canonical_mcp_name() helper in tool_naming.py — a structural check for canonical 
  <server>__<tool> form. The name reflects that this is a shape check, not a registry lookup; any x__y 
 passes. It's still good enough to distinguish MCP calls from harness-internal ones because native tools 
 (Read, Bash, update_topic, ...) never contain the canonical __ separator.
  - `TrajectoryMatcher` now filters native/harness-internal tools (`Read`, `Bash`, `run_shell_command`,
  `update_topic`, `ToolSearch`, Codex's `shell`, ...) from both expected and actual trajectories before
  Jaccard/Levenshtein scoring. Behavior is controlled by a new `filter_native_tools` config flag, **default 
  `true`**; set to `false` when an evalset deliberately scores native-tool usage.
  - Generators (`gemini_cli`, `claude_code`, `codex_cli`) keep returning the full raw tool list from
  `extract_tools()` — filtering is owned by the scorer so raw trajectories stay available for debugging and the
  opt-out is honored end-to-end.
  - Removed the now-redundant `_CLAUDE_INFRA_TOOLS = {"ToolSearch"}` carve-out — subsumed by the generic MCP-only
  filter in the scorer.
  - Updated `datasets/gemini-cli-tools/gemini-cli.evalset.json` expected trajectories to reflect what the agent
  actually invokes (e.g. defensive `get_instance` before `update_instance`).
  - Added `IsCanonicalMcpNameTest` for the helper and 4 new `TrajectoryMatcherTest` cases (default-on filter,
  opt-out flag, symmetric filtering on expected, all-filtered-out edge case).
  - Docs (`gemini_cli_agent_testing.md`, `claude_code_agent_testing.md`, `codex_cli_agent_testing.md`) and example
  `run_config.yaml` files for all three harnesses updated to document the default behavior and how to opt out.

  ## Motivation
  The trajectory matcher uses Jaccard set-similarity and the analyzer only counts a scenario as correct at `score
  == 100`. Native harness tools were leaking into the actual set, dragging Jaccard below 1.0 and producing 0/2
  correctness even when the agent called every MCP tool we cared about. Filtering at the scorer (rather than at
  each generator's `extract_tools()`) lets dataset authors keep `expected_trajectory` focused on user-visible MCP
  intent, preserves raw trajectories for debugging, and gives evalset authors a clean opt-out via
  `filter_native_tools: false` when native-tool usage is part of what they want to measure.

  ## Test plan
  - [x] `pytest evalbench/test/trajectory_matcher_test.py evalbench/test/tool_naming_test.py` — 33/33 pass
  - [x] Re-run a Gemini CLI eval scenario that previously scored 0/2 due to native-tool leakage and confirm it now
  hits 100
  - [x] Run one scenario with `filter_native_tools: false` and confirm native tools appear in the actual set

### DIFF
--- a/datasets/claude-code-tools/example_run_config.yaml
+++ b/datasets/claude-code-tools/example_run_config.yaml
@@ -20,7 +20,13 @@ runners:
 ### Scorer Related Configs
 ############################################################
 scorers:
+  # By default trajectory_matcher drops native/harness-internal tools
+  # (Read, Bash, Edit, ToolSearch, ...) from both expected and actual
+  # trajectories before scoring. Uncomment the block below to keep them
+  # and score native-tool usage too.
   trajectory_matcher: {}
+  # trajectory_matcher:
+  #   filter_native_tools: false
   goal_completion:
     model_config: datasets/model_configs/gemini_2.5_pro_model.yaml
   behavioral_metrics:
--- a/evalbench/generators/models/tool_naming.py
+++ b/evalbench/generators/models/tool_naming.py
@@ -128,13 +128,15 @@ def canonicalize_gemini_tool_name(name: str) -> str:
     return canonical_tool_name(server, tool)
 
 
-def is_canonical_mcp_name(name: str) -> bool:
-    """Return True iff ``name`` is in canonical MCP form (``<server>__<tool>``).
+def looks_like_canonical_mcp_name(name: str) -> bool:
+    """Return True iff ``name`` *looks like* canonical MCP form (``<server>__<tool>``).
 
-    Native/built-in harness tools (Read, Bash, update_topic,
-    run_shell_command, etc.) never contain the canonical separator, so
-    this predicate cleanly distinguishes MCP calls from harness-internal
-    ones after canonicalization.
+    This is a structural check only -- any ``x__y`` with non-empty
+    segments passes; there is no registry of real MCP servers to
+    validate against. Native/built-in harness tools (Read, Bash,
+    update_topic, run_shell_command, etc.) never contain the canonical
+    separator, so the predicate is still good enough to distinguish MCP
+    calls from harness-internal ones after canonicalization.
     """
     if not name:
         return False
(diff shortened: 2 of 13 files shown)


datasets/claude-code-tools/example_run_config.yaml
```diff
@@ -20,7 +20,13 @@ runners:
 ### Scorer Related Configs
 ############################################################
 scorers:
+  # By default trajectory_matcher drops native/harness-internal tools
+  # (Read, Bash, Edit, ToolSearch, ...) from both expected and actual
+  # trajectories before scoring. Uncomment the block below to keep them
+  # and score native-tool usage too.
   trajectory_matcher: {}
+  # trajectory_matcher:
+  #   filter_native_tools: false
```
prernakakkar-google, line 29: Update this in google3 document

evalbench/generators/models/tool_naming.py
```diff
@@ -126,3 +126,17 @@ def canonicalize_gemini_tool_name(name: str) -> str:
         return name
     server, tool = parsed
     return canonical_tool_name(server, tool)
+
+
+def is_canonical_mcp_name(name: str) -> bool:
```
prernakakkar-google: nit looks_like_canonical_mcp_name since it can also pass foo__bar

prernakakkar-google (dismissed):

evalbench/generators/models/tool_naming.py
omkargaikwad23: Good catch — agreed it's a shape check, not a registry check. I think is_canonical_mcp_name parallels 
 stdlib str.isdigit() / Path.is_absolute() (form predicates with is_), and the docstring is explicit that 
 it checks form. Happy to rename if you feel strongly; otherwise I'd lean on the stdlib idiom and the 
 docstring carrying the precision.

prernakakkar-google: because any x__y will pass it does not guarantee that its is_mcp.... So ideally we should rename it

omkargaikwad23: I have renamed the function name to looks_like_canonical_mcp_name. Thanks for the feedback.

prernakakkar-google (approved):

PR #404, opened 2026-05-26
Set up generator class and auth for Data Engineering Agent 

minzznguyen:


IsmailMehdi (commented):
## Code review

Scaffolds a new `DataEngineeringAgentGenerator` plus a `GcpAdcCredentialService` for the A2A SDK's `AuthInterceptor`. Per the WIP label, auth is the focus; messaging in `generate_internal` is stubbed.

### Correctness & functional issues
- **Generator is not registered.** `evalbench/generators/models/__init__.py`'s `get_generator()` dispatch has no `data_engineering_agent` branch and the class isn't imported. Even with a config pointing at it, instantiation falls through to `ValueError("Unknown Generator ...")`. Either add the branch now or call this out as a follow-up in the PR description.
- **Missing dependency declaration.** `a2a` (the A2A SDK) is not in `pyproject.toml`. The repo recently migrated to `uv` (d4501e4) — please `uv add a2a` so the lockfile is updated. Without this, the import fails anywhere outside the author's local env.
- **Blocking sync call inside `async def`.** `credentials.refresh(Request())` does a synchronous HTTPS round-trip and is called from `async def get_credentials`. Inside an asyncio event loop this blocks every other task. Use `await asyncio.to_thread(credentials.refresh, Request())`, or an async-compatible auth flow.
- **Token fetched on every call, no caching.** ADC tokens are valid ~1 hour. `google.auth.default()` + `refresh()` runs on every `get_credentials` invocation — wasteful and slow on the hot path. Cache `credentials` on `self` and only refresh when `not credentials.valid` (the library exposes `credentials.expired` / `valid`).
- **Unused import.** `from a2a.types import a2a_pb2 as pb` — `pb` is never referenced. Drop until `generate_internal` actually needs it.
- **`self.logger` is set but never used.** All output goes through `print(...)`. Pick one — and given the rest of the codebase, that should be `logging` (see `gemini.py`, `claude.py`, `generator.py`).

### Security
- **Logging token bytes, even truncated.** `print(f\"... GCP ADC token: {token[:10]}...\")` writes the first 10 chars of an access token to stdout. Even prefixes are sensitive and end up in CI logs / shared terminals. Replace with `self.logger.debug(\"Retrieved GCP ADC token\")` — no bytes of the token at all.
- **`security_scheme_name` is ignored.** The service hands the same bearer token back regardless of scheme. If multiple schemes are ever registered, this is silently wrong. At minimum add a comment that this provider intentionally only services one scheme; ideally branch (or raise) on unknown names.
- **No error handling around `google.auth.default()` / `refresh()`.** Both can raise (`DefaultCredentialsError`, `RefreshError`). Wrap with try/except and log clearly — saves real debugging time later.

### Style / conventions
- **Emoji-laden `print` statements** (`🔑`, `✅`) don't match the rest of the codebase, which uses standard `logging`. Remove emoji and route through the logger.
- **Trailing whitespace** on the line after `self.logger = logging.getLogger(__name__)`.
- The closest existing pattern to mirror is `claude.py` (concise constructor, single `client` attribute, plain `logging`).

### Tests
None added. Understandable for a WIP that doesn't do anything yet, but at least one unit test that constructs the generator with a fake `CredentialService` (no network call) would catch the registration gap and the unused-import problem in CI. Plan to add tests alongside Task 1.3.

### Recommended order
1. Replace the token-prefix `print` with `logger.debug` — no token bytes.
2. Move `credentials.refresh(Request())` off the event loop (`asyncio.to_thread`) and cache the credentials object.
3. Add `a2a` via `uv add` so lockfile/`pyproject.toml` are updated.
4. Register `DataEngineeringAgentGenerator` in `generators/models/__init__.py` once the class is real — or leave a TODO in the PR body.
5. Remove the unused `a2a_pb2 as pb` import; pick `logging` over `print`.
6. Wrap `google.auth.default()` / `refresh()` with try/except.

### Risk summary
Low blast radius today — the generator can't be selected via config and `generate_internal` raises. Main risks are landing dead/insecure scaffolding into `main`: (a) token-prefix logging, (b) missing dep declaration breaking imports if anything reaches it, (c) blocking sync auth call that will silently hurt throughput once Task 1.3 lands. Address (a) and (c) before this merges even as WIP scaffolding.

evalbench/generators/models/data_engineering_agent.py
```diff
@@ -0,0 +1,88 @@
...
+
+class DataEngineeringAgentGenerator(QueryGenerator):
+    """Data Engineering Agent (DEA) Query Generator using the A2A SDK."""
+
+    def __init__(self, querygenerator_config: Dict[str, Any]):
+        super().__init__(querygenerator_config)
+        self.name = "data_engineering_agent"
+        self.endpoint = querygenerator_config.get(
+            "endpoint",
+            "https://geminidataanalytics.googleapis.com/v1/a2a/projects/bq-dataworkeragent-test/locations/us-west4/agents/dataengineeringagent"
```
IsmailMehdi: no hardcoded endpoints, please have a config with defaults.

evalbench/generators/models/data_engineering_agent.py
```diff
@@ -0,0 +1,88 @@
...
+        )
+        self.target_workspace = querygenerator_config.get(
+            "target_workspace",
+            "projects/bq-dataworkeragent-test/locations/us-west4/repositories/agent_demo_dataform/workspaces/james_test_123"
```
IsmailMehdi: https://github.com/GoogleCloudPlatform/evalbench/pull/404/changes#r3313631524:~:text=no%20hardcoded%20endpoints%2C%20please%20have%20a%20config%20with%20defaults.

IsmailMehdi (changes requested):

evalbench/generators/models/data_engineering_agent.py
```diff
@@ -0,0 +1,97 @@
...
+            self.logger.debug("Retrieved GCP ADC token successfully.")
+            return self.credentials.token
+
+        except (DefaultCredentialsError, RefreshError) as e:
+            self.logger.error(
+                "Failed to retrieve or refresh GCP Application Default "
+                "Credentials: %s",
+                e,
+            )
+            return None
```
graceqi-g: are we able to talk to the endpoint if we fail to get the credentials? if not, let's allow the error propagate to the caller instead of suppressing it here (same for the general exception below). i.e. remove the try-except

evalbench/test/data_engineering_agent_test.py
```diff
@@ -0,0 +1,45 @@
...
+sys.path.append(os.path.dirname(os.path.dirname(os.path.abspath(__file__))))
+
+from generators.models import get_generator
+from generators.models.data_engineering_agent import (
+    DataEngineeringAgentGenerator,
+    GcpAdcCredentialService,
+)
+
+
+def test_data_engineering_agent_generator_setup():
```
graceqi-g, line 17: let's also add tests for the bad paths, i.e. when endpoint and/or target_workspace are missing and ensure the value error is being raised

similarly, let's also try to cover the bad paths in GcpAdcCredentialService.get_credentials

graceqi-g (changes requested):

graceqi-g (dismissed):

IsmailMehdi (commented):
A few correctness concerns inline. See review summary for the full context.

evalbench/generators/models/data_engineering_agent.py
```diff
@@ -0,0 +1,97 @@
...
+    ) -> str | None:
+        if security_scheme_name.lower() not in ("oauth", "oauth2"):
+            raise ValueError(
+                f"GcpAdcCredentialService only services 'oauth' or 'oauth2' "
+                f"schemes, got '{security_scheme_name}'"
+            )
+
+        try:
+            if self.credentials is None:
+                self.credentials, _ = google.auth.default(
```
IsmailMehdi: **Blocking call inside `async def`.** `google.auth.default()` performs blocking I/O (reads ADC files, may hit the metadata server). Calling it directly on the event loop will stall other tasks. Wrap it with `asyncio.to_thread(...)` the same way the `refresh` call below does:

```python
self.credentials, _ = await asyncio.to_thread(
    google.auth.default,
    scopes=["https://www.googleapis.com/auth/cloud-platform"],
)
```

evalbench/generators/models/data_engineering_agent.py
```diff
@@ -0,0 +1,97 @@
...
+        context: ClientCallContext | None,
+    ) -> str | None:
+        if security_scheme_name.lower() not in ("oauth", "oauth2"):
+            raise ValueError(
+                f"GcpAdcCredentialService only services 'oauth' or 'oauth2' "
+                f"schemes, got '{security_scheme_name}'"
+            )
+
+        try:
+            if self.credentials is None:
```
IsmailMehdi: **Race on `self.credentials`.** Two concurrent `get_credentials` calls can both observe `self.credentials is None` and each invoke `google.auth.default()`; similarly two callers can both see `not self.credentials.valid` and refresh in parallel. Guard the init+refresh section with an `asyncio.Lock` (created lazily, since the service may be constructed off-loop) so we initialize/refresh exactly once.

evalbench/generators/models/data_engineering_agent.py
```diff
@@ -0,0 +1,97 @@
...
+
+        self.logger = logging.getLogger(__name__)
+
+        self.auth_interceptor = AuthInterceptor(GcpAdcCredentialService())
+        self.logger.info(
+            "A2A AuthInterceptor successfully configured with "
+            "GcpAdcCredentialService."
+        )
+
+    def generate_internal(self, prompt: str) -> subprocess.CompletedProcess:
```
IsmailMehdi: **Misleading return type.** `-> subprocess.CompletedProcess` looks copy-pasted from `codex_cli.py`, but this generator talks to A2A over HTTP — there is no subprocess. Once Task 1.3 lands the real return shape will be something else. Drop the `import subprocess` and either omit the annotation or use a placeholder (`Any`) so reviewers of the follow-up aren't primed to expect a CLI-shaped result.

The abstract base doesn't enforce a return type, so removing it is safe.

evalbench/generators/models/data_engineering_agent.py
minzznguyen: addressed

evalbench/generators/models/data_engineering_agent.py
minzznguyen: addressed by having an asyncio.Lock

evalbench/generators/models/data_engineering_agent.py
minzznguyen: I changed this to Any, thanks for the comment Medhi!

IsmailMehdi (approved):

evalbench/test/data_engineering_agent_test.py
IsmailMehdi: is this resolved ?

PR #407, opened 2026-06-01
Set up Data Engineering Agent Input

minzznguyen:


IsmailMehdi (approved):

evalbench/dataset/deainput.py
```diff
@@ -0,0 +1,57 @@
...
+        # Extract top-level identification
+        self.id = str(raw_dict.get("id", "-1"))
+        self.job_id = job_id
+        self.trace_id = trace_id
+
+        self.nl_prompt = raw_dict.get("starting_prompt", "")
+
+        # Ensure the stringified payload is ready for the A2A connection
+        self.payload_str = json.dumps(raw_dict)
+        self.payload = self.payload_str
```
graceqi-g, line 22: why do we need self.payload?

minzznguyen: I think it's an implicit requirement by the Protobuf schema. I removed the variable and test it, agentgenwork.py throw 
```
 agentgenwork.py:65] Error processing item: 'EvalDeaRequest' object has no attribute 'payload'  
```

graceqi-g (approved):

PR #408, opened 2026-06-02
feat: add AgyCliGenerator support to evaluator and models, including test suite and configuration datasets

omkargaikwad23:
## Summary
  
  Adds first-class support for the Antigravity (agy) CLI as an agent generator in
  EvalBench, alongside the existing gemini_cli / codex_cli / claude_code harnesses.

## What's included

  - AgyCliGenerator `(evalbench/generators/models/agy_cli.py)`: runs the self-updating
  agy native binary in a sandboxed HOME, configuring model selection (via
  settings.json), auth (host mirror or Secret Manager), MCP servers, and plugin/skill
   installs. Wired into the evaluator and models package.
  - MCP correctness: translates cross-harness configs into agy's schema (`httpUrl →
  serverUrl`) and a fail-fast runtime probe that detects servers attaching with zero
  tools (the silent fall-back-to-gcloud failure).
  - Datasets & configs: real, fake (offline `fake_mcp_server`), and skills-based
  evalsets + example run configs for Cloud SQL tool-calling.
  - GCP project de-hardcoding: project IDs now resolve from `EVAL_GCP_PROJECT_ID` at
  runtime (!ENV in YAML, ${...} expansion in JSON with a fail-fast guard in
  dataset.py), so anyone can target their own project with no file edits.
  - Tooling: pins the agy CLI version in the Docker image; adds tool_naming helper.
  -  Harness env-handling fix (`gemini_cli.py, codex_cli.py, claude_code.py`): changed
  `querygenerator_config.get("env", {})` → `querygenerator_config.get("env") or {}`. The
  old form returned None when a config had an explicit env: key with an empty/null
  value (YAML parses env: to None), causing a crash on the next self.env[...]
  assignment; or {} coerces that to an empty dict.
  - Tests: full `agy_cli_test.py + tool_naming_test.py` suites (skill-clone timeout
  cleanup, MCP config round-trip, probe-log error handling, etc.).
  - Docs: docs/agy_cli_agent_testing.md and AGENTS.md updates.

 ## Test plan

  - pytest `evalbench/test/agy_cli_test.py evalbench/test/tool_naming_test.py`
  - Local run: export `EVAL_GCP_PROJECT_ID=<project>` && export 
  `EVAL_CONFIG=datasets/agy-cli-tools/example_run_config.yaml && ./evalbench/run.sh`
  - Fake/offline run via `example_run_fake_config.yaml` (no GCP access needed)


evalbench_service/Dockerfile
```diff
@@ -19,9 +19,14 @@ RUN \
...
-
-
+  apt-get install -y google-cloud-cli
+
+# Antigravity CLI (agy) -- installer writes a SHA-512-verified binary to the
+# directory passed via --dir. The installer exposes no version pinning, but the
+# binary self-updates in the background; AGY_CLI_DISABLE_AUTO_UPDATE freezes it
+# to the build installed here so a run's agy version stays fixed for the image.
+ENV AGY_CLI_DISABLE_AUTO_UPDATE=true
+RUN curl -fsSL https://antigravity.google/cli/install.sh | sh -s -- --dir /usr/local/bin
```
IsmailMehdi: no, this should be installed dynamically like the other tools

IsmailMehdi (commented):
Per-file review notes. Overall: solid contribution — the agy generator is well-rationalized inline and tests cover the adversarial paths (forged MCP calls, schema-cache staleness, adjacency mispairing). Main blockers: empty PR body, the `env or {}` behavior change in 3 unrelated files isn't called out, and the YAML examples leak the internal `astana-evaluation` project. Inline comments below.

AGENTS.md
```diff
@@ -105,9 +107,12 @@ When evaluating agentic frameworks that leverage external tools (e.g., Gemini CL
 
 | Paradigm | Supported Generators | How it Works |
 |---|---|---|
-| **MCP Servers** | `gemini_cli`, `claude_code`, `codex_cli` | Remote HTTP/SSE or local stdio-based Model Context Protocol servers |
+| **MCP Servers** | `gemini_cli`, `claude_code`, `codex_cli`, `agy_cli` | Remote HTTP/SSE or local stdio-based Model Context Protocol servers |
 | **Extensions** | `gemini_cli`, `claude_code`, `codex_cli` | GitHub-hosted plugin packages installed idempotently via CLI |
-| **Skills** | `gemini_cli`, `claude_code`, `codex_cli` | Local or registry skill packages enabled/linked in the sandboxed environment |
+| **Skills** | `gemini_cli`, `claude_code`, `codex_cli`, `agy_cli` | Local or registry skill packages enabled/linked in the sandboxed environment |
+
```
IsmailMehdi: The dense agy specifics block fits better in `docs/agy_cli_agent_testing.md`. Leave a one-line pointer here — `AGENTS.md` is the architecture overview, not the agy spec, and this block will bit-rot in the wrong place.

datasets/agy-cli-tools/agy-cli-fake.evalset.json
```diff
@@ -0,0 +1,30 @@
...
+  "scenarios": [
+    {
+      "id": "fake-csql-create-instance-success",
+      "starting_prompt": "Create a new Cloud SQL instance named 'my-fake-db' in project 'astana-evaluation'. Use PostgreSQL 17, and set the password to 'password123'. Also use the 'Development' edition preset.",
+      "conversation_plan": "The user wants to create a database. All required parameters are in the starting prompt. The agent should call create_instance and report the success message back.",
+      "expected_trajectory": [
+        "cloud-sql__create_instance"
+      ],
+      "env": {
+        "GOOGLE_CLOUD_PROJECT": "astana-evaluation"
```
IsmailMehdi: `astana-evaluation` is hardcoded as the GCP project. For a public evalset, prefer a sentinel like `your-project-id` so external users can tell at a glance that it needs substitution. Same applies to the non-fake evalset and the model configs.

datasets/agy-cli-tools/agy-cli.evalset.json
```diff
@@ -0,0 +1,33 @@
+{
+  "scenarios": [
+    {
+      "id": "cloud-sql-debug-01",
+      "starting_prompt": "list all instances in project astana-evaluation",
```
IsmailMehdi: Conversation plan grammar nit: "Once all instances are listed if nl2code exist get its state and validate its RUNNABLE" — the simulated user reads this verbatim to drive the dialog, so clearer prose helps reproducibility. Same `astana-evaluation` leakage note as the fake evalset.

datasets/agy-cli-tools/example_run_config.yaml
```diff
@@ -0,0 +1,39 @@
...
+dataset_format: agent-format
+
+# Orchestrator Configuration
+orchestrator: agent
+model_config: datasets/model_configs/agy_cli_model.yaml
+simulated_user_model_config: datasets/model_configs/gemini_2.5_pro_model.yaml
+
+############################################################
+### Scorer Related Configs
+############################################################
```
IsmailMehdi, line 14: The commented-out `trajectory_matcher.filter_native_tools: false` block is a nice in-context hint. Consider copying it to `example_run_fake_config.yaml` for parity — that file has no such comment.

datasets/agy-cli-tools/example_run_fake_config.yaml
```diff
@@ -0,0 +1,33 @@
...
+
+# Orchestrator Configuration
+orchestrator: agent
+model_config: datasets/model_configs/agy_cli_fake_model.yaml
+simulated_user_model_config: datasets/model_configs/gemini_2.5_pro_model.yaml
+
+############################################################
+### Scorer Related Configs
+############################################################
+scorers:
```
IsmailMehdi, line 15: Missing the `trajectory_matcher` filter-native-tools comment that `example_run_config.yaml` carries. Copy it here so both example files document the option.

datasets/agy-cli-tools/example_run_skills_config.yaml
```diff
@@ -0,0 +1,30 @@
...
+dataset_format: agent-format
+
+# Orchestrator Configuration
+orchestrator: agent
+model_config: datasets/model_configs/agy_cli_skills_model.yaml
+simulated_user_model_config: datasets/model_configs/gemini_2.5_pro_model.yaml
+
+############################################################
+### Scorer Related Configs
+############################################################
```
IsmailMehdi, line 14: This config drops `trajectory_matcher` and `parameter_analysis` versus the other two examples. The omission is unexplained — is that because skill-driven runs use a different tool surface? Either restore the scorers or add a one-line comment in the YAML saying why they're left out.

datasets/model_configs/agy_cli_fake_model.yaml
```diff
@@ -0,0 +1,59 @@
...
+  GOOGLE_CLOUD_PROJECT: "astana-evaluation"
+  GOOGLE_CLOUD_LOCATION: "us-central1"
+  GOOGLE_GENAI_USE_VERTEXAI: "true"
+setup:
+  fake_mcp_servers:
+    "cloud-sql":
+      command: "python"
+      args:
+        - "evalbench/util/fake_mcp_server.py"
+        - "--server-name"
```
IsmailMehdi, line 15: The file self-references its own path inside `args:` (`--config datasets/model_configs/agy_cli_fake_model.yaml`). Rename the file and the fake server breaks silently. If the fake MCP server genuinely requires that path, pass it via an env var resolved at runtime, or add a comment at the top of the file noting the filename/content coupling.

datasets/model_configs/agy_cli_model.yaml
```diff
@@ -0,0 +1,36 @@
...
+
+generator: agy_cli
+
+# Model to use. agy has no --model flag and reads no model env var, so the
+# harness writes this into agy's settings.json `model` key. The value must
+# be the exact agy UI label (NOT an API id like "gemini-2.5-pro") -- get it
+# by running `/model` in interactive agy and reading the value it writes to
+# ~/.gemini/antigravity-cli/settings.json. An unrecognized label is silently
+# ignored and agy falls back to its default model. Omit this key to inherit
+# whatever model the host's real settings.json last selected via `/model`.
```
IsmailMehdi: `Gemini 3.1 Pro (High)` is exactly the kind of magic string that rots when the agy UI relabels models. The inline comment is good — keep it. Minor: `headers.X-Goog-User-Project: astana-evaluation` again leaks the internal project; consider a sentinel for the public example.

evalbench/evaluator/agentevaluator.py
```diff
@@ -58,11 +62,11 @@ def evaluate(
         job_id: str,
         run_time: datetime.datetime,
     ):
-        if isinstance(self.generator, (GeminiCliGenerator, ClaudeCodeGenerator, CodexCliGenerator)):
+        if isinstance(self.generator, (GeminiCliGenerator, ClaudeCodeGenerator, CodexCliGenerator, AgyCliGenerator)):
```
IsmailMehdi, line 65: `AgyCliGenerator` is now added to four separate `isinstance(self.generator, (...))` tuples. This is the third CLI added with the same pattern; the next one will be the fourth touch on these four sites. Worth a follow-up to introduce an `is_agent_cli(g)` predicate or an `AgentCliGenerator` base/mixin. Not blocking this PR.

evalbench/generators/models/__init__.py
```diff
@@ -43,6 +44,8 @@ def get_generator(global_models, model_config_path: str, db: DB = None):
             model = ClaudeCodeGenerator(config)
         if config["generator"] == "codex_cli":
             model = CodexCliGenerator(config)
+        if config["generator"] == "agy_cli":
```
IsmailMehdi: Two-line registration is fine for this PR. Side note: the whole file is now a chain of `if config["generator"] == "X"` — a dict dispatch (`{"agy_cli": AgyCliGenerator, ...}`) would be cleaner and harder to typo. Out of scope here.

evalbench/generators/models/agy_cli.py
```diff
@@ -0,0 +1,1076 @@
+from .generator import QueryGenerator
```
IsmailMehdi, line 1: Substantive file — strong on rationale comments (the `serverUrl` vs `httpUrl`, the `gcp.project` load-bearing note, the model-label gotcha all land well). Specific notes:

1. `__init__` does a lot: settings init, auth mirror, ADC handling, MCP/skills setup, verification. Splitting `_setup_auth()` and `_setup_tools()` would help readability.
2. `_run_agy_cli` (~line 1685): bare `except Exception` + `logging.warning` swallows the traceback. Use `logging.exception(...)` so a parser regression surfaces during dev.
3. `_clone_skill_repo`: `git clone --depth 1 --branch <tag>` won't resolve commit SHAs. The doc advertises `#<tag>` pinning generically; document the constraint or fall back to fetch-by-sha.
4. `_install_agy_plugin`: argv has no `--` separator before `target`. A YAML-supplied target starting with `--` would be parsed as a flag. No injection risk (no shell), but worth hardening.
5. `_AGY_SECRET_PATH_RE` accepts `latest` — add a comment that the token is snapshot at init time, so mid-run rotation doesn't take effect for the live process.
6. Logging mixes f-strings (`logging.info(f"Configured MCP server: {server_name}")`) with lazy-format. Standardize on lazy `%s` form (most of the file already does).
7. `AGY_CLI = "agy"` is exported and used as `agent_version` in the evaluator, but the field name suggests a version and it's a bare command name. Either rename or expose `self.agy_cli_version = AGY_CLI` for parity with `gemini_cli_version`/`codex_cli_version`/`claude_code_version`.

evalbench/generators/models/tool_naming.py
```diff
@@ -128,6 +129,73 @@ def canonicalize_gemini_tool_name(name: str) -> str:
...
+    included). Round-tripping through ``json.loads`` strips the quoting;
+    if the value isn't valid JSON we fall back to the raw string.
+    """
+    if isinstance(value, str):
+        try:
+            decoded = json.loads(value)
+            if isinstance(decoded, str):
+                return decoded
+        except (json.JSONDecodeError, ValueError):
+            pass
```
IsmailMehdi, line 163: `_agy_decode_scalar`'s comment ("agy stores each `call_mcp_tool` arg value as a raw JSON token, so a string value arrives JSON-encoded *with* its surrounding quotes") is exactly the kind of weird-thing-future-me-needs-to-know note. Keep it.

Minor: `_AGY_SERVER_KEYS = ("ServerName", "serverName", "server_name", "server")` — accepting four variants is defensive but loose. If only `ServerName` has been observed in v1.0.3, consider narrowing to that and falling back to the others with a warning, so schema drift surfaces early instead of being papered over.

evalbench/test/agy_cli_test.py
```diff
@@ -0,0 +1,995 @@
+import json
```
IsmailMehdi, line 1: Solid coverage and adversarial cases. Patterns to consider:

- Heavy repetition of the 4-line `real_home = tmp_path / "real_home"; real_home.mkdir(); monkeypatch.chdir(...); monkeypatch.setenv("HOME", ...)` setup — a `@pytest.fixture` would shave ~80 lines.
- `_REAL_MCP_CALL` is duplicated as a test fixture here and `REAL_ARGS` in `tool_naming_test.py`. A cross-reference comment between them would let the two stay in lockstep.
- Missing tests: `_translate_mcp_config` direct round-trip; `_verify_mcp_runtime` `OSError` on unreadable `cli.log`; `_clone_skill_repo` timeout cleanup.
- `_local_app_data_dir()` hardcodes `.venv/fake_home_agy` — works because the tests `chdir(tmp_path)`, but the indirection isn't obvious. Worth a one-line comment on the helper.

evalbench/test/tool_naming_test.py
```diff
@@ -124,6 +126,74 @@ def test_gemini_malformed_mcp_returned_as_is(self):
         )
 
 
+class ParseAgyMcpToolCallTest(unittest.TestCase):
```
IsmailMehdi, line 129: Thorough — covers JSON-quoted values (real-world fixture), unquoted fallback, snake_case keys, native passthrough, missing-identity fallback, and the negative case (gemini's `mcp_<server>_<tool>` form must NOT match agy). Good.

evalbench_service/Dockerfile
```diff
@@ -19,9 +19,14 @@ RUN \
...
-  apt-get install -y google-cloud-cli 
-
-
+  apt-get install -y google-cloud-cli
+
+# Antigravity CLI (agy) -- installer writes a SHA-512-verified binary to the
+# directory passed via --dir. The installer exposes no version pinning, but the
+# binary self-updates in the background; AGY_CLI_DISABLE_AUTO_UPDATE freezes it
+# to the build installed here so a run's agy version stays fixed for the image.
+ENV AGY_CLI_DISABLE_AUTO_UPDATE=true
```
IsmailMehdi, line 29: `AGY_CLI_DISABLE_AUTO_UPDATE=true` freeze is the right call. The `curl … | sh` install of agy has no SHA pin of the install script itself — acceptable for an internal tool, but worth flagging since the rest of the Dockerfile uses GPG-keyed apt repos. Consider downloading the script, verifying a known SHA, then executing, or add a one-line comment acknowledging the trust-on-first-use.

datasets/agy-cli-tools/agy-cli.evalset.json
omkargaikwad23: Good catch — fixed. Rewrote the conversation_plan for cloud-sql-debug-01 to clear prose so the simulated user can drive the dialog unambiguously: "Ask the agent to list all instances in project astana-evaluation. Once the instances are listed, if an instance named 'nl2code' exists, get its details and verify that its state is RUNNABLE."

Also reviewed the other conversation_plan/starting_prompt fields across both agy-cli evalsets - no other broken prose. Handling the astana-evaluation leakage note separately.

datasets/agy-cli-tools/example_run_skills_config.yaml
omkargaikwad23: Good catch! Skill-driven runs use a different tool surface where the agent activates high-level skills rather than making direct granular tool calls. I've added an inline comment to the YAML to clarify why they are left out!

evalbench/generators/models/__init__.py
omkargaikwad23: Makes sense! I agree that a dictionary dispatch would be much cleaner and less error-prone here. Since it's out of scope for this PR, I'll leave the two-line registration as-is for now, and we can handle that cleanup in a separate follow-up PR.

AGENTS.md
omkargaikwad23: Makes sense! I have removed that block (already present in agy spec doc) from here and update the skills row.

datasets/agy-cli-tools/agy-cli-fake.evalset.json
omkargaikwad23: Good catch — fixed, and we went a step further than a sentinel. Instead of swapping
   astana-evaluation for a placeholder string that users still have to
  find-and-replace, the project is now pulled from the environment at load time:
  
  - Model configs (agy_cli_model.yaml, agy_cli_skills_model.yaml):
  GOOGLE_CLOUD_PROJECT: !ENV ${EVAL_GCP_PROJECT_ID} (resolved by pyaml_env).
  - Evalsets (agy-cli.evalset.json): use ${EVAL_GCP_PROJECT_ID} in prompts and
  per-scenario env, expanded by a new _expand_env_placeholders helper in dataset.py.

  EVAL_GCP_PROJECT_ID is evalbench's canonical project var (already exported by
  run.sh/cloudbuild/k8s), so external users just export 
  EVAL_GCP_PROJECT_ID=<their-project> and run — no file edits. The loader also fails 
  fast with a clear error if the var is unset, instead of silently passing a literal
  ${...} to the agent. The fake evalset/config use a neutral example-project literal
  since the fake MCP server ignores the value.

datasets/agy-cli-tools/example_run_config.yaml
omkargaikwad23: Added in other run config files. Thanks!

datasets/agy-cli-tools/example_run_fake_config.yaml
omkargaikwad23: Good catch! Added now.

datasets/model_configs/agy_cli_model.yaml
omkargaikwad23: Thanks! Added ${EVAL_GCP_PROJECT_ID} variable to replace actual project id.

evalbench/evaluator/agentevaluator.py
omkargaikwad23: Noted! I will raise a separate PR to introduce an is_agent_cli(g) predicate or an AgentCliGenerator base/mixin.

evalbench_service/Dockerfile
omkargaikwad23: Makes sense for Docker/CI — happy to fetch it at runtime instead of baking it into the image.
  
  One thing I want to flag before I implement: agy can't fully mirror the npm-exec
  pattern. npm exec <cli>@<ver> pulls a pinned version from a registry into npm's own
   cache; agy has no registry, no version pinning, and self-updates. So "dynamic
  install" here can only mean "not baked into the image, fetched on first run" — it
  won't give us the same reproducibility, and whatever version install.sh serves that day is what runs.

  Also, for local runs the install isn't as contained as the npm tools: install.sh
  drops a real binary on the developer's machine (and we'd need it on PATH) rather
  than caching under a tool-owned dir. So I want to check the intent:

  - Option A: install-if-missing everywhere (Docker + local), idempotent — skip if
  agy is already on PATH, one-time download otherwise. Uniform, but does mutate a
  local dev's machine on first run.
  - Option B: auto-install only in the container; on local, just error with the 
  install command if it's missing, so we don't silently install on someone's machine.

  I'd lean B for least surprise locally, but happy to do A if you'd rather keep one
  path. Either way I'll gate on shutil.which so it's a no-op when agy already exists,
   and install to a persistent dir so latency is one-time, not per-run.

omkargaikwad23:
> Per-file review notes. Overall: solid contribution — the agy generator is well-rationalized inline and tests cover the adversarial paths (forged MCP calls, schema-cache staleness, adjacency mispairing). Main blockers: empty PR body, the `env or {}` behavior change in 3 unrelated files isn't called out, and the YAML examples leak the internal `astana-evaluation` project. Inline comments below.

Thanks for the detailed review! I was still refining the code, which is why the PR description was empty initially.
I have now added a full PR description, and I made sure to explicitly call out the env or {} behavior change in the PR body as you requested. I also addressed the gcp project name leakage (by replacing it with runtime environment variable substitution) and have resolved all of your inline comments. Let me know if everything looks good to you now!

evalbench/generators/models/agy_cli.py
omkargaikwad23: Thanks for the feedback. I have addressed all the comments.

evalbench_service/Dockerfile
omkargaikwad23: Good flag. Pinning the script's SHA would break the build every time Google updates
   install.sh (it's served live), so high churn for little gain — unlike apt, there's
   no stable signing model for a one-shot curl. We do still get HTTPS to the Google
  domain plus the installer's own SHA-512 check on the binary; only the script body
  is trust-on-first-use. I added a "Trust model" comment acknowledging that and why
  we accept it for this internal tool.

evalbench/generators/models/tool_naming.py
omkargaikwad23: Good call. I verified the v1.0.3 binary and confirmed it only uses ServerName and ToolName. I've removed the fallback spellings and updated the tests to reject them.

prernakakkar-google:
fix test failures

datasets/model_configs/agy_cli_model.yaml
```diff
@@ -0,0 +1,38 @@
...
+model: "Gemini 3.1 Pro (High)"
```
prernakakkar-google, line 12: How we are planning to name model. This looks specific type of string we are expecting.

evalbench/generators/models/agy_cli.py
```diff
@@ -0,0 +1,1124 @@
...
+        # Authoritative check: each server must have discovered >=1 tool.
+        failed = []
+        loaded = {}
+        for server in configured_servers:
+            server_dir = os.path.join(mcp_schema_root, server)
+            tools = []
+            if os.path.isdir(server_dir):
+                tools = [
+                    f[:-len(".json")]
+                    for f in os.listdir(server_dir)
```
prernakakkar-google: Possible false positive in MCP attach verification: this currently treats any *.json under <appDataDir>/mcp/<server>/ as a discovered tool. If agy ever writes a sidecar JSON file there (or the directory contains junk), we’d incorrectly conclude the server attached successfully. Consider tightening this to validate the JSON contents or at least require a minimal tool-schema shape before counting it as a discovered tool.

datasets/model_configs/agy_cli_model.yaml
```diff
@@ -0,0 +1,38 @@
...
+setup:
+  mcp_servers:
+    "cloud-sql":
+      # agy's MCP schema names the HTTP endpoint field "serverUrl" (the
+      # Windsurf/cortex lineage), not gemini's "httpUrl". The binary's
+      # config struct has no httpUrl field, so an httpUrl value is parsed
+      # to a nil ServerUrl -- no transport, no tools, silent fallback to
+      # gcloud shell-outs. authProviderType/oauth/headers are all native
+      # agy fields, so Google auth works without Bearer-header injection.
+      serverUrl: "https://sqladmin.googleapis.com/mcp"
```
prernakakkar-google, line 30: nit: in other agents we have kept model file generic and parsed while copying settings.json which make model file essentially same across.

datasets/model_configs/agy_cli_skills_model.yaml
```diff
@@ -0,0 +1,33 @@
...
+  CLOUD_SQL_POSTGRES_PROJECT: !ENV ${EVAL_GCP_PROJECT_ID}
+  CLOUD_SQL_POSTGRES_REGION: "us-central1"
+  CLOUD_SQL_POSTGRES_INSTANCE: <instance>
+  CLOUD_SQL_POSTGRES_DATABASE: <database>
+  CLOUD_SQL_POSTGRES_USER: <user>
+  CLOUD_SQL_POSTGRES_PASSWORD: <password>
+  CLOUD_SQL_POSTGRES_IP_TYPE: "PUBLIC"
+setup:
+  # `skills` is named for parity with the claude_code/codex_cli harnesses.
+  # For agy each entry installs a *plugin* (`agy plugin install`), which may
```
prernakakkar-google, line 26: I think we should support both skill and plugin

prernakakkar-google (changes requested):

datasets/model_configs/agy_cli_model.yaml
omkargaikwad23: Please check the comment written above model flag.

evalbench/test/agy_cli_test.py
omkargaikwad23: Good catch! I have added the @pytest.fixture, synced the cross-reference comments, added the missing tests, and documented the helper. Thanks for the thorough review!

datasets/model_configs/agy_cli_fake_model.yaml
omkargaikwad23: Done. Added the 3-line coupling-warning comment to the top. The identical self-reference exists in claude_code_fake_model.yaml, gemini_cli_fake_model.yaml, and codex_cli_fake_model.yaml.

evalbench_service/Dockerfile
IsmailMehdi: we discussed using the session directory for the install t avoid collisions, let me know if that makes sense.

datasets/model_configs/agy_cli_model.yaml
prernakakkar-google: We would want to provide model selection to user so we should be very clear what we are expecting here. We will not always want default model.

omkargaikwad23: That's exactly why I added the detailed comment block above it. Since agy forces us to use these specific UI label strings instead of standard API IDs, the user will need to update this field in their YAML config if they want to test a different model. The comment serves as the guide for how they can find the exact string they need. I am making sure to update to document this so users see it easily!

evalbench/generators/models/agy_cli.py
omkargaikwad23: Good catch — added defense-in-depth. _verify_mcp_runtime now counts a *.json only if it parses as a JSON object with a non-empty name (new _is_tool_schema_file helper). In practice the dir is wiped before the probe and agy writes only tool schemas (verified against the real cloud-sql cache — all 16 pass), so this just guards against agy emitting a sidecar file in future or a partial write. Added tests for the junk cases.

datasets/model_configs/agy_cli_skills_model.yaml
omkargaikwad23: Understood. I'll handle this in a follow-up PR.

datasets/model_configs/agy_cli_model.yaml
omkargaikwad23: While httpUrl also works here because our harness parses and translates it, I explicitly used serverUrl because it is the native field that Agy uses. My thought was that exposing the native serverUrl in the example would be less confusing for users who are already familiar with Agy's internals.

prernakakkar-google:
token count seems to be 0. Do you plan to address it?

omkargaikwad23:
> token count seems to be 0. Do you plan to address it?

Currently, the Agy CLI does not log LLM token consumption to disk. Because of this, I have temporarily commented out the token_consumption scorer. We will need to wait for Agy to expose usage metrics upstream before we can fully support it here.

prernakakkar-google:
> > token count seems to be 0. Do you plan to address it?
> 
> Currently, the Agy CLI does not log LLM token consumption to disk. Because of this, I have temporarily commented out the token_consumption scorer. We will need to wait for Agy to expose usage metrics upstream before we can fully support it here.

/context gave me this:
────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────
└ Context Usage
◉ ◉ ◉ ◉ ◉ ◉ ◉ ◉ ◉ □ □ □ □ □ □ □ □ □ □ □ □ □ □ □     Gemini 3.5 Flash (Medium) · 27.2k/1.0M tokens
□ □ □ □ □ □ □ □ □ □ □ □ □ □ □ □ □ □ □ □ □ □ □ □      (2.6%)
□ □ □ □ □ □ □ □ □ □ □ □ □ □ □ □ □ □ □ □ □ □ □ □     Token usage by category
□ □ □ □ □ □ □ □ □ □ □ □ □ □ □ □ □ □ □ □ □ □ □ □     ◉ User messages: 19 tokens (0.0%)
□ □ □ □ □ □ □ □ □ □ □ □ □ □ □ □ □ □ □ □ □ □ □ □     ◉ Agent responses: 4.5k tokens (0.4%)
□ □ □ □ □ □ □ □ □ □ □ □ □ □ □ □ □ □ □ □ □ □ □ □     ◉ Tool calls: 648 tokens (0.1%)
□ □ □ □ □ □ □ □ □ □ □ □ □ □ □ □ □ □ □ □ □ □ □ □     ⛁ System prompt: 5.9k tokens (0.6%)
□ □ □ □ □ □ □ □ □ □ □ □ □ □ □ □ □ □ □ □ □ □ □ □     ⛁ System tools: 15.1k tokens (1.4%)
□ □ □ □ □ □ □ □ □ □ □ □ □ □ □ □ □ □ □ □ □ □ □ □     ⛁ Skills: 453 tokens (0.0%)
□ □ □ □ □ □ □ □ □ □ □ □ □ □ □ □ □ □ □ □ □ □ □ □     ⛁ Subagents: 654 tokens (0.1%)
□ □ □ □ □ □ □ □ □ □ □ □ □ □ □ □ □ □ □ □ □ □ □ □     □ Free space: 1.0M (97.4%)
□ □ □ □ □ □ □ □ □ □ □ □ □ □ □ □ □ □ □ □ □ □ □ □     ⊠ Checkpoint buffer: 306 tokens (not counted in usage)


Checkpoints (1) · /rewind
└ Checkpoint 1 (active, in context): Verifying Workspace Context
The user's main objective is to test an...

System files · auto-loaded
 └ ~/.gemini/antigravity/skills/

prernakakkar-google:
As per https://paste.googleplex.com/5270759760068608 stats does not show which model was used for call. Lets address what model it is instead of "agy"

omkargaikwad23:
> > > token count seems to be 0. Do you plan to address it?
> > 
> > 
> > Currently, the Agy CLI does not log LLM token consumption to disk. Because of this, I have temporarily commented out the token_consumption scorer. We will need to wait for Agy to expose usage metrics upstream before we can fully support it here.
> 
> /context gave me this: ──────────────────────────────────────────────────────────────────────────────────────────────────────────────────────── └ Context Usage ◉ ◉ ◉ ◉ ◉ ◉ ◉ ◉ ◉ □ □ □ □ □ □ □ □ □ □ □ □ □ □ □ Gemini 3.5 Flash (Medium) · 27.2k/1.0M tokens □ □ □ □ □ □ □ □ □ □ □ □ □ □ □ □ □ □ □ □ □ □ □ □ (2.6%) □ □ □ □ □ □ □ □ □ □ □ □ □ □ □ □ □ □ □ □ □ □ □ □ Token usage by category □ □ □ □ □ □ □ □ □ □ □ □ □ □ □ □ □ □ □ □ □ □ □ □ ◉ User messages: 19 tokens (0.0%) □ □ □ □ □ □ □ □ □ □ □ □ □ □ □ □ □ □ □ □ □ □ □ □ ◉ Agent responses: 4.5k tokens (0.4%) □ □ □ □ □ □ □ □ □ □ □ □ □ □ □ □ □ □ □ □ □ □ □ □ ◉ Tool calls: 648 tokens (0.1%) □ □ □ □ □ □ □ □ □ □ □ □ □ □ □ □ □ □ □ □ □ □ □ □ ⛁ System prompt: 5.9k tokens (0.6%) □ □ □ □ □ □ □ □ □ □ □ □ □ □ □ □ □ □ □ □ □ □ □ □ ⛁ System tools: 15.1k tokens (1.4%) □ □ □ □ □ □ □ □ □ □ □ □ □ □ □ □ □ □ □ □ □ □ □ □ ⛁ Skills: 453 tokens (0.0%) □ □ □ □ □ □ □ □ □ □ □ □ □ □ □ □ □ □ □ □ □ □ □ □ ⛁ Subagents: 654 tokens (0.1%) □ □ □ □ □ □ □ □ □ □ □ □ □ □ □ □ □ □ □ □ □ □ □ □ □ Free space: 1.0M (97.4%) □ □ □ □ □ □ □ □ □ □ □ □ □ □ □ □ □ □ □ □ □ □ □ □ ⊠ Checkpoint buffer: 306 tokens (not counted in usage)
> 
> Checkpoints (1) · /rewind └ Checkpoint 1 (active, in context): Verifying Workspace Context The user's main objective is to test an...
> 
> System files · auto-loaded └ ~/.gemini/antigravity/skills/

Thanks for sharing that /context output! However, because our harness executes agy in headless mode (agy -p), interactive slash commands unfortunately don't execute. I'll ask the agy team how we can get these metrics programmatically instead. Could we unblock this PR for now and handle the token tracking as a fast-follow?

evalbench_service/Dockerfile
omkargaikwad23: Thanks Mehdi for the suggestion. We are installing the binary in sandbox now at runtime.

omkargaikwad23:
> As per https://paste.googleplex.com/5270759760068608 stats does not show which model was used for call. Lets address what model it is instead of "agy"

Thanks for pointing it out. I have addressed the default model name in the PR instead of showing "agy". Now, we are doing regex search `cli.log`file to get the model name if default is not provided .

IsmailMehdi (approved):

prernakakkar-google (approved):

PR #415, opened 2026-06-05
refactor: introduce AgentCliGenerator base to unify CLI generator handling

omkargaikwad23:
## Summary
  - Add `AgentCliGenerator` base class (ABC) that the four CLI generators
    (gemini_cli, claude_code, codex_cli, agy_cli) now inherit from, declaring
    the shared interface: `version`, `create_command`, `safe_generate`,
    `parse_response`, `extract_tools`, `extract_skills`.
  - Unify `create_command` to one signature across all four generators so the
    evaluator no longer branches on concrete type.
  - Simplify `AgentEvaluator`: collapse the four `isinstance(...4-tuple...)`
    checks to a single `isinstance(self.generator, AgentCliGenerator)`, replace
    the per-type version chain with `self.generator.version`, and drop all
    concrete CLI imports.

  ## Why
  Each new CLI generator previously required edits at ~6 sites in the evaluator.
  Keying off the base class makes adding a generator zero-touch in the evaluator,
  and the ABC enforces the interface at instantiation.

  ## Test plan
  - [x] `pytest test/agy_cli_test.py test/gemini_cli_test.py` — 52 passed
  - [x] All four subclasses report no remaining abstract methods
  - [x] `agentevaluator.py` imports cleanly, no concrete CLI references
  - [x] Dockerised run logs: https://paste.googleplex.com/4915247398912000

### DIFF
--- a/evalbench/evaluator/agentevaluator.py
+++ b/evalbench/evaluator/agentevaluator.py
@@ -56,7 +56,7 @@ class AgentEvaluator:
             return self._evaluate_agent_cli(dataset, job_id, run_time)
         else:
             raise NotImplementedError(
-                "This evaluator currently only supports AgentCliGenerator subclasses")
+                "This evaluator currently only supports GeminiCliGenerator, ClaudeCodeGenerator, CodexCliGenerator and AgyCliGenerator")
 
     def _evaluate_agent_cli(
         self,
--- a/evalbench/evaluator/agentevaluator.py
+++ b/evalbench/evaluator/agentevaluator.py
@@ -7,10 +7,7 @@ import threading
 
 from dataset.evalgeminicliinput import EvalGeminiCliRequest
 from generators.models import get_generator
-from generators.models.gemini_cli import GeminiCliGenerator
-from generators.models.claude_code import ClaudeCodeGenerator
-from generators.models.codex_cli import CodexCliGenerator
-from generators.models.agy_cli import AgyCliGenerator
+from generators.models.agent_cli import AgentCliGenerator
 from mp import mprunner
 from work.agentgenwork import AgentGenWork
 from evaluator.simulateduser import SimulatedUser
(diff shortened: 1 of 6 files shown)


prernakakkar-google:
Please manually test against docker and attach results here.

evalbench/evaluator/agentevaluator.py
```diff
@@ -62,11 +52,11 @@ def evaluate(
         job_id: str,
         run_time: datetime.datetime,
     ):
-        if isinstance(self.generator, (GeminiCliGenerator, ClaudeCodeGenerator, CodexCliGenerator, AgyCliGenerator)):
+        if isinstance(self.generator, AgentCliGenerator):
             return self._evaluate_agent_cli(dataset, job_id, run_time)
         else:
             raise NotImplementedError(
-                "This evaluator currently only supports GeminiCliGenerator, ClaudeCodeGenerator, CodexCliGenerator and AgyCliGenerator")
+                "This evaluator currently only supports AgentCliGenerator subclasses")
```
prernakakkar-google: I think lets keep previous error message here

prernakakkar-google (changes requested):

evalbench/generators/models/agy_cli.py
```diff
@@ -1234,7 +1238,7 @@ def safe_generate(
 
     def create_command(
         self, cli: str, prompt: str, env: dict = None, resume: bool = False,
-        cwd: str = None,
+        session_id: str = None, cwd: str = None,
```
IsmailMehdi, line 1241: at some point, we should support cwd for these generators.

IsmailMehdi (dismissed):

evalbench/evaluator/agentevaluator.py
omkargaikwad23: The main goal of this PR is to make AgentEvaluator completely agnostic to the specific generators so we don't have to edit this file every time we add a new one. If we hardcode the generator names in this error message, it will become outdated as soon as a new harness is added (or it will force us to keep updating agentevaluator.py). Since the type checker now enforces AgentCliGenerator, would you be okay with keeping the generic message to avoid this future maintenance overhead?

prernakakkar-google: Mentioned this because as a user I am not aware what is AgentCliGenerator subclass is. I dont expect any new harness to be added anytime soon.

omkargaikwad23: Makes sense! Added back the prev comment.

IsmailMehdi (approved):

omkargaikwad23:
> Please manually test against docker and attach results here.

Added. Please check.

prernakakkar-google (approved):

PR #420, opened 2026-06-08
Add tool calls and responses to conversation history

silviojr:


IsmailMehdi:
Do we still need this PR ?

IsmailMehdi:
is this WIP? what are we doing with this one ?

PR #422, opened 2026-06-09
fix(dea): resolve concurrency deadlock in GcpAdcCredentialService

minzznguyen:
### Fix design
We use `threading.Lock` instead of `asyncio.Lock` because:
- when using parallel runner agents in config, the agents run across threads
- if we use `asyncio.Lock` it would create deadlock because agents cannot coordinate or wake up tasks waiting in a different event loop on another thread
- threading.Lock solve this exact problem

### DIFF
--- a/evalbench/generators/models/gcp_data_engineering_agent.py
+++ b/evalbench/generators/models/gcp_data_engineering_agent.py
@@ -1,5 +1,6 @@
 import asyncio
 import logging
+import threading
 from typing import Any
 
 from a2a.client import ClientCallContext
@@ -18,12 +19,13 @@ class GcpAdcCredentialService(CredentialService):
     """GCP Application Default Credentials (ADC) service for A2A SDK.
 
     This provider only services OAuth/OAuth2 schemes.
+    Thread-safe and Loop-safe implementation utilizing standard threading.Lock
+    and a fast-path check to avoid thread pool overhead for valid tokens.
     """
 
     def __init__(self):
-        self.logger = logging.getLogger(__name__)
         self.credentials = None
-        self._lock = None
+        self._lock = threading.Lock()
 
     async def get_credentials(
         self,
@@ -36,40 +38,41 @@ class GcpAdcCredentialService(CredentialService):
                 f"schemes, got '{security_scheme_name}'"
             )
 
-        if self._lock is None:
-            self._lock = asyncio.Lock()
-
         try:
-            async with self._lock:
-                if self.credentials is None:
-                    credentials, _ = await asyncio.to_thread(
-                        google.auth.default,
-                        scopes=[
-                            "https://www.googleapis.com/auth/cloud-platform"
-                        ]
-                    )
-                    self.credentials = credentials
-
-                if not self.credentials.valid:
-                    await asyncio.to_thread(
-                        self.credentials.refresh, Request()
-                    )
-
-                self.logger.debug("Retrieved GCP ADC token successfully.")
-                return self.credentials.token
-
+            return await asyncio.to_thread(self._get_and_refresh_token)
         except (DefaultCredentialsError, RefreshError) as e:
-            self.logger.error(
-                "Failed to retrieve or refresh GCP Application Default "
-                "Credentials: %s",
-                e,
-            )
+            logger.error("GCP ADC authentication failed: %s", e)
             raise
         except Exception as e:
-            self.logger.exception(
-                "Unexpected error while fetching GCP ADC credentials: %s", e
+            logger.exception(
+                "Unexpected error while retrieving GCP ADC credentials: %s", e
             )
             raise
+
+    def _get_and_refresh_token(self) -> str:
+        with self._lock:
+            if self.credentials is None:
+                logger.debug(
+                    "Initializing GCP Application Default Credentials."
+                )
+                credentials, _ = google.auth.default(
+                    scopes=["https://www.googleapis.com/auth/cloud-platform"]
+                )
+                self.credentials = credentials
+
+            if not self.credentials.valid:
+                logger.debug(
+                    "GCP ADC token is invalid or expired. Refreshing..."
+                )
+                self.credentials.refresh(Request())
+
+            token_val = self.credentials.token
+            if not token_val:
+                raise ValueError(
+                    "GCP ADC token is empty after retrieval/refresh."
+                )
+
+            return token_val
 
 
 class DataEngineeringAgentGenerator(QueryGenerator):
--- a/evalbench/test/gcp_data_engineering_agent_test.py
+++ b/evalbench/test/gcp_data_engineering_agent_test.py
@@ -1,5 +1,8 @@
+import asyncio
 import os
 import sys
+import threading
+import time
 from unittest.mock import MagicMock, patch
 import pytest
 from google.auth.exceptions import DefaultCredentialsError, RefreshError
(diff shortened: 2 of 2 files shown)


IsmailMehdi (commented):
Concurrency fix and async-loop rationale make sense — `asyncio.Lock` is bound to the event loop it was created in, so this would deadlock (or `RuntimeError`) under the parallel runner. A few things to address inline before merging; the most important is the racy lazy lock init in `_get_and_refresh_token`, which actually undermines the fix under the workload this PR targets.

evalbench/generators/models/gcp_data_engineering_agent.py
```diff
@@ -17,7 +18,8 @@
 class GcpAdcCredentialService(CredentialService):
     """GCP Application Default Credentials (ADC) service for A2A SDK.
 
-    This provider only services OAuth/OAuth2 schemes.
+    Thread-safe and Loop-safe implementation utilizing standard threading.Lock
+    and a fast-path check to avoid thread pool overhead for valid tokens.
```
IsmailMehdi, line 23: Nit: the previous docstring noted *"This provider only services OAuth/OAuth2 schemes."* That contract is still enforced at line 35 and worth keeping in the docstring alongside the new thread-safety note.

evalbench/generators/models/gcp_data_engineering_agent.py
```diff
@@ -36,41 +38,40 @@ async def get_credentials(
                 f"schemes, got '{security_scheme_name}'"
             )
 
-        if self._lock is None:
-            self._lock = asyncio.Lock()
+        # Fast path: return valid token immediately without thread hop or lock.
+        creds = self.credentials
+        if creds is not None and creds.valid:
```
IsmailMehdi: Minor TOCTOU: this reads `creds.valid` outside the lock while another thread may be inside `creds.refresh(Request())` mutating the same object. In practice the GIL + `google.auth.Credentials` make this benign, but it depends on internals of a third-party library. Safer pattern: store a snapshot `self._cached_token: str | None` that's only assigned inside the lock after a successful refresh, and have the fast path read only that. Then no shared object is concurrently read+mutated.

evalbench/generators/models/gcp_data_engineering_agent.py
```diff
@@ -36,41 +38,40 @@ async def get_credentials(
...
+            token = creds.token
+            if token is not None:
```
IsmailMehdi: Inconsistent with L70 (`if not self.credentials.token`). Here `token = ""` passes through; there it raises `ValueError`. Pick one — `if not token:` here matches the locked-path check and rejects empty strings.

evalbench/generators/models/gcp_data_engineering_agent.py
```diff
@@ -36,41 +38,40 @@ async def get_credentials(
...
-                "Credentials: %s",
-                e,
-            )
-            raise
+            return await asyncio.to_thread(self._get_and_refresh_token)
         except Exception as e:
-            self.logger.exception(
-                "Unexpected error while fetching GCP ADC credentials: %s", e
-            )
+            self.logger.error("Failed to retrieve GCP ADC credentials: %s", e)
```
IsmailMehdi: Regression vs. the previous code: the old version used `self.logger.exception(...)` for unexpected errors, which captures the traceback. This `self.logger.error(...)` only logs `str(e)`. For unexpected failures the traceback is the only thing that makes them debuggable — use `self.logger.exception(...)` or pass `exc_info=True`. Also worth keeping the `(DefaultCredentialsError, RefreshError)` arm distinct from the catch-all so expected auth failures aren't logged at the same loudness as bugs.

evalbench/generators/models/gcp_data_engineering_agent.py
```diff
@@ -36,41 +38,40 @@ async def get_credentials(
...
             raise
 
+    def _get_and_refresh_token(self) -> str:
+        if self._lock is None:
+            self._lock = threading.Lock()
```
IsmailMehdi: **Blocker.** This lazy init is not thread-safe and defeats the purpose of the PR. Two threads entering `_get_and_refresh_token` for the first time can both see `self._lock is None`, each construct their own `threading.Lock()`, and the second assignment clobbers the first — leaving the two threads holding *different* lock objects. The critical section below is then no longer mutually exclusive, which is exactly the scenario this PR is meant to fix.

Fix: initialize eagerly in `__init__` with `self._lock = threading.Lock()` and delete L55–56. There's no reason to defer construction.

evalbench/generators/models/gcp_data_engineering_agent.py
```diff
@@ -36,41 +38,40 @@ async def get_credentials(
...
+        with self._lock:
+            # Double-Checked Locking: Check again inside the lock
+            if self.credentials is None:
+                self.logger.info("Initializing GCP Application Default Credentials.")
```
IsmailMehdi: Nit: bumped from `.debug` to `.info`. Init only fires once per service instance, so this is fine, but worth a sanity check on log volume if multiple `GcpAdcCredentialService` instances are constructed per run.

evalbench/generators/models/gcp_data_engineering_agent.py
```diff
@@ -36,41 +38,40 @@ async def get_credentials(
...
+                credentials, _ = google.auth.default(
+                    scopes=["https://www.googleapis.com/auth/cloud-platform"]
+                )
+                self.credentials = credentials
+
+            if not self.credentials.valid:
+                self.logger.info("GCP ADC token is invalid or expired. Refreshing...")
```
IsmailMehdi: Same nit as L60 — refresh fires roughly once per hour per credential, so `.info` is probably fine, just flagging the level change from the previous `.debug`.

evalbench/generators/models/gcp_data_engineering_agent.py
```diff
@@ -36,41 +38,40 @@ async def get_credentials(
...
+                self.credentials.refresh(Request())
+
+            if not self.credentials.token:
```
IsmailMehdi: See L45 — this check (`not self.credentials.token`) rejects both `None` and `""`, but the fast path's `if token is not None` only rejects `None`. Align them.

evalbench/generators/models/gcp_data_engineering_agent.py
```diff
@@ -36,41 +38,40 @@ async def get_credentials(
...
+                raise ValueError("GCP ADC token is empty after retrieval/refresh.")
+
+            return self.credentials.token
```
IsmailMehdi: No test added for the regression being fixed. A concurrency bug worth a PR is worth a test that would fail without the patch — a small `ThreadPoolExecutor` running N concurrent `get_credentials` calls with `google.auth.default` mocked to sleep briefly would catch both the original deadlock and the L56 race noted above. Otherwise the next refactor reintroduces the problem with no signal.

evalbench/generators/models/gcp_data_engineering_agent.py
minzznguyen: addressed

evalbench/generators/models/gcp_data_engineering_agent.py
minzznguyen: I took a look at this and seems like having fast path is not straightforward, I'm reverting back to having everything inside the lock since the time optimized with fast path is negligible

evalbench/generators/models/gcp_data_engineering_agent.py
minzznguyen: addressed

evalbench/generators/models/gcp_data_engineering_agent.py
minzznguyen: addressed

evalbench/generators/models/gcp_data_engineering_agent.py
minzznguyen: addressed

evalbench/generators/models/gcp_data_engineering_agent.py
minzznguyen: addressed by changing back to debug just to be safe

evalbench/generators/models/gcp_data_engineering_agent.py
minzznguyen: addressed by changing back to debug just to be safe

evalbench/generators/models/gcp_data_engineering_agent.py
minzznguyen: addressed

IsmailMehdi (approved):
remove WIP and we can merge this

PR #427, opened 2026-06-12
Implement Cloud Dataform Scorer

minzznguyen:
### Description
Design a scorer for Evalbench to evaluate Dataform projects that reside in the cloud (remote repositories/workspaces), as the current dataformscorer.py only supports local filesystems.

Use the Google Cloud Dataform API directly from the Python scorer to trigger compilation and execution in the cloud.

### DIFF
--- a/datasets/dea-tools/README.md
+++ b/datasets/dea-tools/README.md
@@ -6,7 +6,8 @@ This directory contains a sample configuration for evaluating the Data Engineeri
 The model config is located at `datasets/model_configs/gcp_data_engineering_agent_model.yaml`. It uses environment variables to dynamically resolve your GCP coordinates and target Dataform workspace:
 *   `gcp_project_id`: Read from `EVAL_GCP_PROJECT_ID`
 *   `gcp_region`: Read from `EVAL_GCP_PROJECT_REGION` (defaults to `us-west4` if unset)
-*   `target_workspace`: Read from `EVAL_DEA_WORKSPACE`
+*   `dataform_repository_id`: Read from `EVAL_DEA_REPOSITORY_ID`
+*   `dataform_workspace_id`: Read from `EVAL_DEA_WORKSPACE_ID`
 
 ## 2. Supply Your Evaluation Dataset
 The dataset file is defined in `datasets/dea-tools/dea-live-conversational.evalset.json`. It defines conversational turns (such as reading table schemas and modifying schemas) along with evaluation metrics and rubrics.
@@ -19,16 +20,16 @@ To run the evaluation, make sure you are in the root directory of the `evalbench
 PROTOCOL_BUFFERS_PYTHON_IMPLEMENTATION=python \
 EVAL_GCP_PROJECT_ID=<YOUR_GCP_PROJECT_ID> \
 EVAL_GCP_PROJECT_REGION=<YOUR_GCP_REGION> \
-EVAL_DEA_REPOSITORY=<YOUR_REPO_NAME> \
-EVAL_DEA_WORKSPACE=<YOUR_WORKSPACE_NAME> \
+EVAL_DEA_REPOSITORY_ID=<YOUR_REPO_ID> \
+EVAL_DEA_WORKSPACE_ID=<YOUR_WORKSPACE_ID> \
 .venv/bin/python3 evalbench/evalbench.py --experiment_config=datasets/dea-tools/example_run_config.yaml
 ```
 
 ### Key Environment Variables:
 *   `EVAL_GCP_PROJECT_ID`: The GCP Project ID where your DEA agent is deployed.
 *   `EVAL_GCP_PROJECT_REGION`: The GCP Region (e.g., `us-west4`) of the agent.
-*   `EVAL_DEA_REPOSITORY`: The target Dataform repository name.
-*   `EVAL_DEA_WORKSPACE`: The target Dataform workspace name.
+*   `EVAL_DEA_REPOSITORY_ID`: The target Dataform repository ID (short name).
+*   `EVAL_DEA_WORKSPACE_ID`: The target Dataform workspace ID (short name).
 
 ## 4. Inspect Results
 Upon completion, results will be generated under the `results/` folder:
--- a/evalbench/scorers/dataformcloudscorer.py
+++ b/evalbench/scorers/dataformcloudscorer.py
@@ -1,0 +1,236 @@
+"""Cloud Dataform Scorers using Google Cloud Dataform API."""
+
+import time
+from typing import Tuple, List, Any
+from google.cloud import dataform_v1beta1
+import google.auth
+from google.api_core import exceptions as api_exceptions
+from scorers import comparator
+
+
+class DataformCloudBaseScorer(comparator.Comparator):
+    """Base class for Cloud Dataform scorers using Dataform API."""
+
+    def __init__(self, config: dict):
+        super().__init__(config)
+        self.config = config
+
+        project = config.get("gcp_project_id")
+        location = config.get("gcp_region")
+        repository = config.get("dataform_repository")
+        workspace = config.get("dataform_workspace")
+
+        if not project:
+            raise ValueError(
+                "Configuration key 'gcp_project_id' is required for "
+                "DataformCloudBaseScorer."
+            )
+        if not location:
+            raise ValueError(
+                "Configuration key 'gcp_region' is required for "
+                "DataformCloudBaseScorer."
+            )
+        if not repository:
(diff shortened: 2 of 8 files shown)


evalbench/scorers/dataformcloudscorer.py
```diff
@@ -0,0 +1,234 @@
...
+
+    def run_dataform_cloud_command(
+        self, command: List[str]
+    ) -> Tuple[float, str]:
+        """Executes a Dataform cloud command.
+
+        Returns a score and analysis.
+        """
+        try:
+
```
graceqi-g, line 116: nit: remove extra new line

evalbench/scorers/dataformcloudscorer.py
```diff
@@ -0,0 +1,234 @@
...
+            # Step 2: Trigger compilation in the cloud (always required for
```
graceqi-g: can you clean up the comments in this method? e.g. seems step 1 is missing, you can remove the step #s

graceqi-g (dismissed):

evalbench/scorers/dataformcloudscorer.py
minzznguyen: addressed

evalbench/scorers/dataformcloudscorer.py
minzznguyen: addressed

datasets/dea-tools/README.md
```diff
@@ -19,14 +19,16 @@ To run the evaluation, make sure you are in the root directory of the `evalbench
...
+EVAL_DEA_REPOSITORY=<YOUR_REPO_NAME> \
+EVAL_DEA_WORKSPACE=<YOUR_WORKSPACE_NAME> \
 .venv/bin/python3 evalbench/evalbench.py --experiment_config=datasets/dea-tools/example_run_config.yaml
 ```
 
 ### Key Environment Variables:
 *   `EVAL_GCP_PROJECT_ID`: The GCP Project ID where your DEA agent is deployed.
 *   `EVAL_GCP_PROJECT_REGION`: The GCP Region (e.g., `us-west4`) of the agent.
-*   `EVAL_DEA_WORKSPACE`: The fully qualified resource path to the Dataform workspace where the agent will execute actions.
+*   `EVAL_DEA_REPOSITORY`: The target Dataform repository name.
```
saurabh-net: this should say id in the description, same as above.

datasets/dea-tools/README.md
```diff
@@ -19,14 +19,16 @@ To run the evaluation, make sure you are in the root directory of the `evalbench
 PROTOCOL_BUFFERS_PYTHON_IMPLEMENTATION=python \
 EVAL_GCP_PROJECT_ID=<YOUR_GCP_PROJECT_ID> \
 EVAL_GCP_PROJECT_REGION=<YOUR_GCP_REGION> \
-EVAL_DEA_WORKSPACE="projects/<YOUR_GCP_PROJECT_ID>/locations/<YOUR_GCP_REGION>/repositories/<YOUR_REPO_NAME>/workspaces/<YOUR_WORKSPACE_NAME>" \
+EVAL_DEA_REPOSITORY=<YOUR_REPO_NAME> \
```
saurabh-net: Consider changing this to 
EVAL_DEA_REPOSITORY_ID=<REPOSITORY_ID> \
EVAL_DEA_WORKSPACE_ID=<WORKSPACE_ID>

Name typically refers to the full resource path. And ID to the unique identifier.

It tends to cause confusion otherwise.

evalbench/scorers/dataformcloudscorer.py
```diff
@@ -0,0 +1,231 @@
...
+
+        # Setup client
+        self.credentials, _ = google.auth.default()
+        self.client = dataform_v1beta1.DataformClient(
+            credentials=self.credentials
+        )
+
+    def _run_workflow_invocation(
+        self, compilation_result_name: str
+    ) -> Tuple[Any, str, list[str]]:
```
saurabh-net, line 59: out of curiousity, why is this Any instead of dataform_v1beta1.WorkflowInvocationAction.State?

evalbench/scorers/score.py
```diff
@@ -162,6 +163,32 @@ def compare(
         comparators.append(
             dataformscorer.DataformRunScorer(scorers["dataform_run"])
         )
+    if "dataform_cloud_compile" in scorers:
+        scorer_config = scorers["dataform_cloud_compile"].copy()
+        for k in [
+            "gcp_project_id",
```
IsmailMehdi: You are assuming these are the same as what scorer should use. I'd like to see this extraction in the scorer rather than in score.py

evalbench/scorers/dataformcloudscorer.py
```diff
@@ -0,0 +1,231 @@
...
+        while state in [
+            dataform_v1beta1.WorkflowInvocation.State.RUNNING,
+            dataform_v1beta1.WorkflowInvocation.State.CANCELING,
+        ]:
+            if time.time() - start_time > self.timeout_seconds:
+                raise TimeoutError(
+                    f"Workflow invocation timed out after "
+                    f"{self.timeout_seconds} seconds."
+                )
+            time.sleep(5)
```
IsmailMehdi, line 83: this needs to be a config

evalbench/scorers/dataformcloudscorer.py
```diff
@@ -0,0 +1,231 @@
...
+    def run_dataform_cloud_command(
+        self, command: List[str]
+    ) -> Tuple[float, str]:
+        """Executes a Dataform cloud command.
+
+        Returns a score and analysis.
+        """
+        try:
+            # Trigger compilation in the cloud (always required for both compile and run)
+            compilation_result = self.client.create_compilation_result(
```
IsmailMehdi, line 118: When both `dataform_cloud_compile` and `dataform_cloud_run` are enabled — as in the new `example_run_config.yaml` in this PR — each scorer independently calls `create_compilation_result`, so every eval item compiles in the cloud twice. Either have the run scorer accept a previously-computed compilation result (e.g., cache on the instance keyed by `(workspace_name, current_revision)`), or at minimum document that running both is redundant API spend.

evalbench/evaluator/dataengineeringagentevaluator.py
```diff
@@ -42,6 +42,7 @@ def __init__(self, config: dict[str, Any]) -> None:
             model_config = loaded_config.copy()
             model_config.update(config)
 
+        self.config = model_config
```
IsmailMehdi: This line silently changes what `self.config` means downstream: it's now the merged `model_config` (yaml + input overrides) instead of the raw input `config`. That's exactly what makes `score.py`'s `experiment_config.get("gcp_project_id")` (in the new cloud-scorer registration block at score.py:163-185) actually find the GCP keys — but the coupling is invisible. A future refactor that moves or removes this assignment will silently break cloud scorers with `ValueError("'gcp_project_id' is required")`. A short comment here noting the scorer-config dependency, plus an integration test that asserts the scorer receives `gcp_project_id`, would lock the contract in.

datasets/dea-tools/README.md
minzznguyen: yup, that makes sense. addressed!

datasets/dea-tools/README.md
minzznguyen: That's a good catch! I applied changes to add ID for these

evalbench/scorers/dataformcloudscorer.py
minzznguyen: `dataform_v1beta1.WorkflowInvocationAction.State` is a dynamically generated class, which can create some type-checking issues for static type checkers. I use `Any` for simplicity's sake

evalbench/scorers/score.py
minzznguyen: I have addressed this by requiring configs for the scorer be specified in yaml config file

evalbench/scorers/dataformcloudscorer.py
minzznguyen: addressed

evalbench/scorers/dataformcloudscorer.py
graceqi-g: I think we can combine the scores (so cloud dataform scorer does run and compile), but if we do this I would like to update the local dataform scorer to do the same, which would require us to make dashboard updates - for now let's just document that we are doing redundant calls. I'll open a bug for this and we can address it later.

graceqi-g (approved):

PR #436, opened 2026-06-18
build: add Python version classifiers to pyproject.toml

omkargaikwad23:
The package declares requires-python = ">=3.10" but had no "Programming Language :: Python" classifiers, leaving the supported versions undiscoverable on PyPI. Add classifiers for 3.10-3.13 to match the requires-python constraint.


wangauone (approved):
All good! Thanks Omkar!

PR #437, opened 2026-06-18
fix: add timeout to data agent request in querydata generator

omkargaikwad23:
The requests.post call to the ADK API server had no timeout, so a stalled or unresponsive server would block the evaluation indefinitely. Add a bounded (connect, read) timeout to fail fast instead of hanging.


wangauone (dismissed):
Thanks Omkar! I bet we also need to add the same timout in https://github.com/GoogleCloudPlatform/evalbench/blob/main/evalbench/generators/models/query_data_api.py#L60C36-L60C46 here. I think we can do in the same PR?

omkargaikwad23:
> Thanks Omkar! I bet we also need to add the same timout in https://github.com/GoogleCloudPlatform/evalbench/blob/main/evalbench/generators/models/query_data_api.py#L60C36-L60C46 here. I think we can do in the same PR?

Makes sense! I have updated this in the same PR. Thanks!

prernakakkar-google (approved):

PR #453, opened 2026-06-22
Feat: Add ability to filter to specific json scenario/examples

g-lynnzee:
Avoid having to clobber json files when trying to test a specific scenario (which is also token intensive by agents).

Usage example:

./evalbench/evalbench.py \
  --experiment_config=core-cujs/run_gemini_cli.yaml \
  --scenarios=autoctx-expert-eval-confirm


IsmailMehdi:
PTAL codestyle and pytest failures

g-lynnzee:
PTAL

IsmailMehdi (approved):

PR #456, opened 2026-06-23
fix(dea): accumulate all recursion agent text parts

minzznguyen:
### Description
- old implementation was only able to extract the top level of the agent's response
- old implementation only output unnecessary logs when finding iterable objects
- this change provide fix by appending message text instead of returning prematurely, and have helper functions for finding iterable objects

### DIFF
--- a/evalbench/generators/models/gcp_data_engineering_agent.py
+++ b/evalbench/generators/models/gcp_data_engineering_agent.py
@@ -120,6 +120,17 @@ class GcpAdcCredentialService(CredentialService):
             return token_val
 
 
+def _is_iterable(obj: Any) -> bool:
+    """Safely checks if an object is iterable, excluding strings and bytes."""
+    if isinstance(obj, (str, bytes)):
+        return False
+    try:
+        iter(obj)
+        return True
+    except TypeError:
+        return False
+
+
 def _extract_message_text(msg: Any) -> str:
     """Extracts agent message text from a single Message object."""
     if getattr(msg, "role", None) != pb.ROLE_AGENT:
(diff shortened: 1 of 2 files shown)


evalbench/generators/models/gcp_data_engineering_agent.py
```diff
@@ -336,7 +345,7 @@ async def _run_client(
             message_req.metadata[CONVERSATION_TOKEN_URI] = token
 
         context = ClientCallContext(
-            timeout=180.0,
+            timeout=600.0,
```
graceqi-g: any reason you decided to increase the timeout? 10 minutes seems like a long time to be waiting for a response. if you need to increase it, maybe try 5 min to start

graceqi-g (changes requested):
can you attach log of the parsed agent text before and after your change?

also, do the UTs for the generator need to be updated? can you add some around the text parsing if possible?

evalbench/generators/models/gcp_data_engineering_agent.py
IsmailMehdi: good point

minzznguyen: Make sense, I will make the time out to be 5min. The reason I did this is because when I test more complex cases like NYC Data vault it tends to take more time

IsmailMehdi (dismissed):

minzznguyen:
> can you attach log of the parsed agent text before and after your change?

Here is the formatted before and after log - key point to look is that the before section only has messages saying what the agent will do for each turn, while the after section contains messages saying what the agent will do and also message reviewing the code that the agent just generated:

https://docs.google.com/document/d/1WfbFvipzvzBXdex4j8OZHzZwl7eqihq1md4rLKIVL1M/edit?tab=t.jh4qb22b4ubk

graceqi-g:
> > can you attach log of the parsed agent text before and after your change?
> 
> Here is the formatted before and after log - key point to look is that the before section only has messages saying what the agent will do for each turn, while the after section contains messages saying what the agent will do and also message reviewing the code that the agent just generated:
> 
> https://docs.google.com/document/d/1WfbFvipzvzBXdex4j8OZHzZwl7eqihq1md4rLKIVL1M/edit?tab=t.jh4qb22b4ubk

looks good! are you able to add a UT to the generator to ensure the agent text is parsed properly?

evalbench/generators/models/gcp_data_engineering_agent.py
minzznguyen: addressed second point by adding UTs to make sure generator parsed sucessfully

graceqi-g (approved):

PR #459, opened 2026-06-25
feat(mcp-readability): add MCP style-guide compliance eval

akangsha7:
Add an `mcp_readability` orchestrator to EvalBench that checks MCP endpoints against the Data Cloud MCP style guide and reports compliance metrics (P0/P1/P2 counts, tool count, token budget usage, score).

- Tool fetching uses the official `mcp` SDK over Streamable HTTP (`streamable_http_client` + `ClientSession.list_tools`) with URL sanitization and ADC auth; a `file` source supports offline runs.
- Tools are rendered to man-page markup (resolves $ref/$defs, flattens nesting, renders enum/default) which feeds the LLM scorer.
- Scorer uses Vertex Gemini JSON mode, anchors severity to style-guide priority annotations, supports per-endpoint waivers, evaluates from an LLM-agent-consumption perspective, and runs a dedicated consistency second pass against prior feedback when available.
- Results write to a compliance CSV and, when `reporting.bigquery` is configured, append to the existing eval BigQuery table.
- Endpoint list mirrors monitored_endpoints.textproto (24 endpoints).

### DIFF
--- a/datasets/mcp_readability/sample_tools.yaml
+++ b/datasets/mcp_readability/sample_tools.yaml
@@ -1,0 +1,37 @@
+# Sample MCP tools spec for offline, deterministic testing of the readability
+# check (tools_source.type: file). Intentionally mixes good and bad style so the
+# LLM has something to flag.
+tools:
+  - name: list_datasets
+    description: "List all BigQuery datasets in the given project. Use this to discover available datasets before querying."
+    inputSchema:
+      type: object
+      properties:
+        project_id:
+          type: string
+          description: "The Google Cloud project ID to list datasets from."
+      required: ["project_id"]
+
+  - name: RunQuery
+    description: "runs sql"
+    inputSchema:
+      type: object
+      properties:
+        q:
+          type: string
+
+  - name: delete_table
+    description: "Deletes a table."
+    inputSchema:
+      type: object
+      properties:
+        project_id:
+          type: string
+          description: "Project ID."
+        dataset_id:
+          type: string
+          description: "Dataset ID."
+        table_id:
+          type: string
+          description: "Table ID."
+      required: ["project_id", "dataset_id", "table_id"]
--- a/datasets/mcp_readability/style_guide.md
+++ b/datasets/mcp_readability/style_guide.md
@@ -1,0 +1,699 @@
+# Data Cloud MCP Tool Style Guide
+
+[TOC]
+
+*   **Go link:**
+    [go/mcp-datacloud-styleguide](http://go/mcp-datacloud-styleguide)
+
+## Style Principles
+
+This document provides style guidelines and best practices for Data Cloud teams
+building MCP (Model Context Protocol) tools.
+
+> [!NOTE] The tool descriptions and examples in this guide are generated using
+> the `experimental/users/kvg/mcp_readability` utility. You can use this tool to
+> view the API of your own MCP tools in a readable format by running it against
+> your MCP server endpoint:
+>
+> ```bash
+> blaze run //experimental/users/kvg/mcp_readability:mcp_list_tools -- \
+>     --url="https://your-mcp-server.googleapis.com/mcp" \
+>     --auth \
+>     --project="your-gcp-project"
+> ```
+>
+> You can use the following tool to get LLM generated recommendations (AI
+> evaluation) based on this style guide for your MCP tools by running it against
+> your MCP server endpoint. This tool requires a Gemini API key to run the AI
+> evaluation:
+>
+> ```bash
+> blaze run cloud/databases/mcp/readability:run_mcp_readability_check -- \
+>     --gemini_api_key <gemini_api_key> check_url \
+>     --url "https://your-mcp-server.googleapis.com/mcp" \
+>     [--tools <tool_names>] [--auth] [--project <project_id>]
+> ```
+
+There are a few overarching principles that summarize how to think about
+building good MCP tools. The following are attributes of good tools, in order of
+importance:
+
+1.  **Clarity**: The tool's purpose, parameters, and error messages are clear to
+    the reasoning engine.
+2.  **Simplicity**: Tools and their parameters are simple enough to reduce
+    cognitive load on the model, minimize hallucination risks, and avoid
+    excessive token usage, while remaining comprehensive enough to complete a
+    task.
+3.  **Security**: Tools are secure by default, do not expose secrets, and use
+    granular protections.
+4.  **Maintainability**: Tools are built in a consistent manner, relying on
+    standardized platforms, making them easier to maintain across teams.
+
+## Platform & Constraints
+
+### Platform to build tools <!-- priority: p0 -->
+
+All teams within data cloud **MUST** be using either OneMCP or MCP Toolbox for
+tool building. Bespoke solutions are **NOT** supported.
+
+### Tool Count Limits <!-- priority: p0 -->
+
+The current rule of thumb is to keep it to 40 tools per server as an upper
+limit, though even this amount may cause performance issues. Performance
+degrades as more tools are added, so teams **SHOULD** heavily weigh adding new
+tools against the negative impact on tool accuracy until other mechanisms are in
+place to deal with this.
+
+In the future, we aspire for this number to be reduced to 5-8 tools per server
+as toolsets/skills capabilities improve in MCP Toolbox and OneMCP, though this
+is not a requirement today. Until overall platform tool handling improves, teams
+**SHOULD** minimize tool count as much as possible.
+
+## Tool Design & API
+
+MCP tools **MUST** be high quality, reliable, and easy for agents to use.
+
+### Tool Names <!-- priority: p1 for <action>_<resource>, p2 for snake_case -->
+
+Teams **MUST** use `snake_case` (`<action>_<resource>`) for tool names. You
+**SHOULD NOT** use product-specific prefixes, as the agent can automatically
+disambiguate by prefixing the tool name with the MCP server name.
+
+```text {.bad}
+================================================================================
+TOOL: my_service_create_instance
+================================================================================
+================================================================================
+TOOL: my_service_execute_sql
+================================================================================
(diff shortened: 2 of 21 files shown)


evalbench/scorers/mcp_style_compliance.py
```diff
@@ -0,0 +1,360 @@
...
+        """Generate the model response as raw JSON text.
+
+        Prefer Gemini's native JSON mode via the underlying genai client, which
+        guarantees syntactically valid JSON and -- crucially -- bypasses
+        ``GeminiGenerator.generate``'s SQL sanitizer (it strips backslashes and
+        collapses whitespace, corrupting JSON escapes). Falls back to the
+        generic ``generate`` for non-Gemini models.
+        """
+        client = getattr(self.model, "client", None)
+        caller = getattr(self.model, "_call_generate_content", None)
```
IsmailMehdi, line 226: lets not do this, either add a flag or a new signature to skip the sanitizer instead of jamming into the internals

evalbench/evaluator/mcp_readability/enums.py
```diff
@@ -0,0 +1,89 @@
...
+    """
+
+    CHECK_STATUS_UNSPECIFIED = 0
+    SUCCESS = 1
+    FETCH_ERROR = 2
+    ANALYSIS_ERROR = 3
+    INTERNAL_ERROR = 4
+
+
+def _coerce(enum_cls, value, default):
```
IsmailMehdi, line 57: this seems unecessary, we control the config don't we ?

evalbench/generators/models/mcp_tools.py
```diff
@@ -0,0 +1,229 @@
...
+    # mcp source (official SDK over Streamable HTTP)
+    # ------------------------------------------------------------------
+    def _from_mcp(self, source: dict, endpoint: dict) -> list[mcp_types.Tool]:
+        raw_url = source.get("url") or endpoint.get("endpoint_url")
+        if not raw_url:
+            raise McpToolsError("tools_source.type 'mcp' requires an endpoint URL")
+        url = self.sanitize_url(raw_url)
+        headers = self._auth_headers(source)
+        try:
+            return anyio.run(
```
IsmailMehdi, line 152: `anyio.run(...)` creates a fresh event loop. That's fine today because the orchestrator drives this from its `concurrent.futures.ThreadPoolExecutor` (each worker thread has no running loop). But if this generator is ever called from an existing async context — e.g., the agent-flow async refactors that have been landing — it will raise `RuntimeError: cannot be called from running event loop`, and the orchestrator will record a `FETCH_ERROR` row with a confusing message.

Two options:
- Detect an existing loop and use `anyio.from_thread.run(...)` in that case, so the generator works in either context.
- Or add an explicit comment here that this generator is sync-only by design and will break if invoked from async code, so the next caller doesn't get surprised.

evalbench/evaluator/mcp_readability/orchestrator.py
```diff
@@ -0,0 +1,334 @@
...
+                row["check_status"] = CheckStatus.ANALYSIS_ERROR.name
+                row["error_message"] = f"{type(e).__name__}: {e}"
+                return row
+
+            # 5. Success: the eval ran. Compliance findings are recorded as
+            #    metrics independent of this status.
+            row.update(
+                {
+                    "check_status": CheckStatus.SUCCESS.name,
+                    "p0_issues": int(feedback.get("p0_issues", 0)),
```
IsmailMehdi, line 240: If the LLM returns a non-numeric value here (e.g. `"p0_issues": "five"` or `null` in a way that survives `_parse`), `int(...)` raises `ValueError` / `TypeError`. The outer `try` does catch it — but it lands in the generic `except Exception` block below as `INTERNAL_ERROR` rather than the more accurate `ANALYSIS_ERROR` (the model's output is the actual problem). The scorer already exposes `_safe_int` for exactly this; lift it (or re-define alongside) and use it for all four fields:

```python
from scorers.mcp_style_compliance import _safe_int
...
"p0_issues": _safe_int(feedback.get("p0_issues", 0)),
"p1_issues": _safe_int(feedback.get("p1_issues", 0)),
"p2_issues": _safe_int(feedback.get("p2_issues", 0)),
"compliance_score": _safe_int(feedback.get("compliance_score", 0)),
```

Keeps a malformed model response on the `ANALYSIS_ERROR` path with full row metrics still attempted, rather than dropping to `INTERNAL_ERROR` with no compliance fields.

evalbench/generators/models/mcp_tools.py
```diff
@@ -0,0 +1,229 @@
...
+        self.name = "mcp_tools"
+        cfg = querygenerator_config or {}
+        self.timeout = cfg.get("timeout", 30)
+        self.default_headers = cfg.get("headers") or {}
+        # Optional default auth mode applied when an endpoint doesn't specify one.
+        self.default_auth = cfg.get("auth")
+
+    # The abstract base requires generate_internal; the orchestrator calls
+    # fetch_tools directly, but we keep this so the class is a valid generator.
+    def generate_internal(self, prompt):
```
IsmailMehdi, line 53: This override is a soft-fail shim — `""` for any non-dict input — that mostly exists to satisfy the abstract base. The orchestrator never calls it (it uses `fetch_tools` directly).

Problem: if a future caller plumbs this generator into the standard generation path and passes a string prompt, they get back an empty string with no log and no error — the failure looks like the generator succeeded with no content. That's a debugging nightmare.

Two cleaner options:
- Drop the override and let the base class's abstract method raise.
- Or fail loud:
  ```python
  def generate_internal(self, prompt):
      raise NotImplementedError(
          "McpToolsGenerator does not support text generation; "
          "call fetch_tools(endpoint, defaults) directly."
      )
  ```

The second option is friendliest because the error message tells the next caller exactly what to do.

evalbench/evaluator/mcp_readability/orchestrator.py
```diff
@@ -0,0 +1,334 @@
...
+        self.environment_filter = {str(e).strip().upper() for e in env_filter}
+
+        # Previous feedback for run-to-run consistency.
+        self.previous_feedback = self._load_previous_feedback(
+            config.get("previous_results_csv")
+        )
+
+        # Shared model registry for get_generator (lock + cache), as used by
+        # the standard orchestrators.
+        self.global_models = {
```
IsmailMehdi, line 84: This is a fresh per-orchestrator `global_models` dict. As a result, `get_generator(self.global_models, ...)` always misses the registry and instantiates a new model client (Gemini, etc.) per orchestrator. If multiple orchestrators run in the same process (a future suite-config scenario, or someone wiring eval_service to construct several in sequence), they each re-build the same model rather than reusing one.

The other orchestrators accept `global_models` as a constructor argument so callers can share it. Suggest matching that pattern here:

```python
def __init__(self, config, db_configs, setup_config, report_progress=False, global_models=None):
    super().__init__(config, db_configs, setup_config, report_progress)
    ...
    self.global_models = global_models or {
        "lock": threading.Lock(),
        "registered_models": {},
    }
```

Low blast radius (kwarg default preserves current behavior) and keeps the door open to model reuse. Minor — small N of endpoints today, but it's the kind of thing that ossifies into "why does this re-init?" later.

IsmailMehdi:
PTAL failed actions and CLA

IsmailMehdi (changes requested):

datasets/mcp_readability/style_guide.md
helloeve, line 1: can we revert this change?

evalbench/evalbench.py
```diff
@@ -137,6 +142,17 @@ def eval(experiment_config: str):
...
+            # This orchestrator writes its own compliance CSV in process() and
+            # intentionally returns no NL2SQL results/scores temp files.
+            logging.info(
+                f"Finished MCP readability run for config '{display_config}'."
+            )
+            reporters = []
+            config_df = None
+            results_df = None
+            scores_df = None
+            summary_scores_df = None
```
helloeve, line 155: can we have the results needed managed within those data frames and reuse the reporter to handle the result upload?

evalbench/evaluator/mcp_readability/orchestrator.py
```diff
@@ -0,0 +1,334 @@
...
+            logging.exception("mcp_readability: failed to write compliance CSV")
+
+        # Optionally append the same rows to the existing eval BigQuery table.
+        if reporting.get("bigquery"):
+            self._write_bigquery(reporting["bigquery"])
+
+        # Return Nones so the standard NL2SQL reporting path is skipped.
+        return (self.job_id, self.run_time, None, None, None)
+
+    def _write_bigquery(self, bq_config):
```
helloeve, line 148: I would recommend to use reporter to handle the result upload rather than the orchestrator doing the work -- this customized path technically works, but introduces additional burden to maintain as it deviates from the normal evalbench flow.

datasets/mcp_readability/sample_tools.yaml
helloeve, line 1: IIUC, this refers to the offline file-based approach; I think we should stick to raw json as input, otherwise it remains unclear how this file can be realistically constructed by a developer and aligned with the man page expected by the eval.

evalbench/evaluator/mcp_readability/orchestrator.py
```diff
@@ -0,0 +1,334 @@
...
+            return ""
+        try:
+            with open(path, "r", encoding="utf-8") as f:
+                return f.read()
+        except OSError as e:
+            logging.warning("mcp_readability: could not read style guide %s: %s", path, e)
+            return ""
+
+    @staticmethod
+    def _load_previous_feedback(csv_path) -> dict:
```
helloeve, line 313: This seems to be relying on local files and I assume we are planning to extend this to read from BQ? My recommendation would be to do this as a follow-up when we get the full pipeline working e2e (results displayed in dashboard). In other words the initial version by default does not have previous feedback, wdyt?

PR #461, opened 2026-06-26
## Fix Sandbox Isolation Violation (cwd omission) in Codex CLI Generator

silviojr:
### Problem
When the Codex CLI generator executed commands (`codex exec`), it did not forward the working directory (`cwd`) parameter to the underlying `subprocess.Popen` invocation. As a result, Codex CLI executed commands and produced side-effects (such as generating files and notebook outputs) in the host project's root directory instead of within the isolated sandbox directory (`.venv/fake_home_codex`).

### Comparison with Other Generators
Other CLI-based generators like `GeminiCliGenerator` and `ClaudeCodeGenerator` correctly accept the `cwd` argument in their commands and forward it to `subprocess.run`/`subprocess.Popen` (defaulting to their respective sandboxed fake home directories). The `CodexCliGenerator` was an outlier failing to do so.

### Changes Made
1. **Added `cwd` parameter**: Added support for the `cwd` parameter to `CLICommand` and `CodexCliGenerator.create_command`.
2. **Passed `cwd` to `subprocess.Popen`**: Updated `CodexCliGenerator._execute_cli_command` to receive the `cwd` argument and pass it to `subprocess.Popen`, defaulting to `self.fake_home` if not specified.
3. **Ignored local testing configs**: Updated `.gitignore` to exclude the entire `datasets/bqml/` folder from version control.

### Tests
Created a new unit test suite [codex_cli_test.py](file:///usr/local/google/home/silviojr/git/evalbench/evalbench/test/codex_cli_test.py) validating that:
* The CLI command correctly runs inside the custom `cwd` when one is explicitly provided.
* The CLI command defaults to executing inside the generator's `self.fake_home` directory when `cwd` is omitted.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -196,11 +196,11 @@ viewer/version.txt
 .jetskicli/project.json
 
 # Local testing datasets/configs (do not commit)
-datasets/bqml/Bullseye_PCA.json
-datasets/bqml/Bullseye_PCA_config.yaml
+datasets/bqml/
 datasets/model_configs/gemini_cli_test_model.yaml
 datasets/model_configs/gemini_2.5_pro_test_model.yaml
 datasets/model_configs/gemini_3_flash_test_model.yaml
+datasets/model_configs/codex_cli_dak_model.yaml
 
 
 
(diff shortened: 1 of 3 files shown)


IsmailMehdi (changes requested):
Please consider the CodeQL changes, otherwise LGTM

silviojr:
thanks. fixed them.

IsmailMehdi (approved):

PR #462, opened 2026-06-26
feat: Improve isolation of agent evaluation CUJs by staging the work_dir to a new tmp directory tree

g-lynnzee:
I ran into this where the agent is able to access and thus will use files in the working_dirs for other CUJs. 

I have a test case that relies on the agent trying and failing to find a dataset file, and acting a certain way. However, the agent sometimes will read other working_dirs and act based on datasets files there.


evalbench/evaluator/agentevaluator.py
```diff
@@ -205,6 +207,8 @@ def process_scenario(
                 else:
                     break
 
+        self._cleanup_sandbox(resolved_work_dir, temp_sandbox_dir)
```
IsmailMehdi, line 210: This cleanup call isn't reached on exception. The turn-loop body has its own `try/except` for `safe_generate` / `generate`, but any exception that bypasses those (e.g. an uncaught error in `extract_tools` / `extract_skills`, `simulated_user.get_next_response`, `KeyboardInterrupt`, or a bug somewhere else in the loop body) skips this line and leaves the tempdir on disk. On a long-running eval server (or a flaky scenario that fails repeatedly), this leaks `evalbench-sandbox-*` directories under `/tmp` indefinitely.

Wrap the post-setup block in `try/finally`:

```python
resolved_work_dir = scenario.get("resolved_work_dir")
execution_cwd, temp_sandbox_dir = self._setup_sandbox(resolved_work_dir)
try:
    # fake_home prep + turn loop + _finalize_scenario all live here
    ...
finally:
    self._cleanup_sandbox(resolved_work_dir, temp_sandbox_dir)
```

Worth a unit test alongside the change: patch `self.generator.safe_generate` to raise, call `process_scenario`, then assert no `evalbench-sandbox-*` directories remain under `tempfile.gettempdir()`. That locks the contract in.

evalbench/evaluator/agentevaluator.py
```diff
@@ -267,3 +271,41 @@ def _finalize_scenario(
...
+    def _setup_sandbox(self, resolved_work_dir: str | None) -> tuple[str | None, str | None]:
+        """Creates a temporary isolated directory and copies resolved_work_dir into it."""
+        if not resolved_work_dir:
+            return None, None
+        os.makedirs(resolved_work_dir, exist_ok=True)
+        temp_sandbox_dir = tempfile.mkdtemp(prefix="evalbench-sandbox-")
+        shutil.copytree(resolved_work_dir, temp_sandbox_dir, dirs_exist_ok=True)
+        return temp_sandbox_dir, temp_sandbox_dir
+
+    def _register_trusted_folders(self, fake_home: str, execution_cwd: str | None, resolved_work_dir: str | None) -> None:
```
IsmailMehdi, line 284: This seems to only work for gemini, we should support others and restrict this one to gemini.

evalbench/evaluator/agentevaluator.py
```diff
@@ -267,3 +271,41 @@ def _finalize_scenario(
...
+        """Writes trusted folder configuration mappings to settings json to prevent CLI trust hangs."""
+        if not execution_cwd:
+            return
+        trusted_folders_path = os.path.join(fake_home, ".gemini", "trustedFolders.json")
+        os.makedirs(os.path.dirname(trusted_folders_path), exist_ok=True)
```
IsmailMehdi, line 289: worth looking at mktemp ?

IsmailMehdi:
PTAL https://github.com/GoogleCloudPlatform/evalbench/pull/461 also

prernakakkar-google:
@g-lynnzee is this still an active PR to be worked on? Should we close this if its not required now?

PR #463, opened 2026-06-28
feat(mcp-readability): add MCP tool man-page formatter

akangsha7:
Render MCP tools (tools/list) into a human-readable "man page" for the readability scorer's LLM judge. Resolves $ref/$defs, flattens nested params into dotted paths, marks REQUIRED/OPTIONAL, renders enum/default, and guards against recursive schemas.

Includes unit tests covering empty/no-param tools, required/optional, enum+default, nested objects, $ref/$defs, arrays of objects, and the recursion guard.

### DIFF
--- a/evalbench/generators/models/mcp_tool_formatter.py
+++ b/evalbench/generators/models/mcp_tool_formatter.py
@@ -1,0 +1,267 @@
+"""Render MCP tools into a human-readable "man page".
+
+This is the *rendering* half of the MCP readability check. It takes a sequence of
+MCP tools (as returned by ``tools/list``) and produces a single man-page style
+string that the readability scorer feeds to an LLM judge.
+
+The renderer:
+  - resolves JSON-Schema ``$ref`` / ``$defs`` references,
+  - flattens nested parameters into dotted paths,
+  - marks each parameter ``REQUIRED`` / ``OPTIONAL``,
+  - renders structured ``enum`` and ``default`` values so style rules that key
+    off them (e.g. "Use Enums") are judgeable from the markup,
+  - guards against infinite recursion in self-referential schemas.
+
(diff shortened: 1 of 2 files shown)


evalbench/generators/models/mcp_tool_formatter.py
```diff
@@ -0,0 +1,279 @@
...
+            )
+    while wrapped and not wrapped[-1]:
+        wrapped.pop()
+    return wrapped
+
+
+def _value_indent(level: int) -> str:
+    """Indent used for sub-lines (description/enum/default) at a given level."""
+    if level == 0:
+        return "      "
```
helloeve: nit: using `" " * num` for better readability in this function?

evalbench/generators/models/mcp_tool_formatter.py
```diff
@@ -0,0 +1,279 @@
...
+        if desc:
+            lines.extend(_wrap_text(desc, indent + "  "))
+    return lines
+
+
+def _format_recursive(level: int) -> list[str]:
+    """Formats a recursive reference notice for array items at the given level."""
+    desc = "[Recursive Reference]"
+    if level == 0:
+        return _wrap_text(desc, "      ")
```
helloeve: nit: consider having this ident defined as shared variable to it remains consistent?

helloeve (commented):
Re ACL: you can clean the commit history so it does not have claude code as part of the author

evalbench/generators/models/mcp_tool_formatter.py
akangsha7: done

IsmailMehdi (approved):

helloeve (approved):

PR #465, opened 2026-06-28
feat: Support cross-database evaluation with SQLite ground truth

arieljassan:
### Overview
This PR adds support for running the BIRD benchmark (and similar cross-database SQL evaluations) on BigQuery while using local SQLite databases for ground truth reference execution. 
When evaluating AI-generated queries on BigQuery against reference queries written in SQLite syntax (e.g., using `STRFTIME` or SQLite math functions), BigQuery cannot execute the reference queries directly. This update introduces a hybrid bridging mechanism that dynamically resolves reference answers from local SQLite files while evaluating generated queries on BigQuery.
### Key Changes
* **`evalbench/databases/bigquery.py`**: Implements the required `ensure_database_exists` abstract method on `BQDB` to fulfill the base `DB` class contract.
* **`evalbench/scorers/sqlite_bridge.py` & `llmrater.py`**: Adds conditional ground truth resolution. When `golden_error` occurs and a hybrid judge is configured, `LLMRater` automatically fetches the true reference execution rows from the local SQLite database.
* **`hybrid_xa_judge.py`**: Adds a self-contained Execution Accuracy (XA) evaluator script for `PythonScorer`. It normalizes across engine data types (`Decimal`/`Int64` vs `float`), ignores column header differences, and compares rows order-independently.
### Verification
* Performed functional end-to-end benchmark tests across dual BIRD evaluation datasets (`california_schools` and `card_games`).
* Confirmed functional ground truth resolution from local SQLite reference tables during live BigQuery evaluation runs.
* Verified clean telemetry data warehousing into BigQuery (`FLOAT64` score compatibility).


IsmailMehdi (commented):
Solves a real problem (cross-engine BIRD eval on BQ), but a few correctness issues to address. Inline comments cover the file-specific items. Three cross-cutting points that don't fit a specific line:

**No tests for any of the new code.** Both `sqlite_bridge.py` and `hybrid_xa_judge.py` are trivially unit-testable (in-memory SQLite via `sqlite3.connect(":memory:")`, mock `sys.argv`). Without tests the fragile bits flagged inline will silently rot. Minimum I'd want to see: `test_get_sqlite_ground_truth_uses_named_db` (after the #2 fix), `test_compare_result_sets_handles_decimal_vs_float`, `test_compare_result_sets_ignores_column_names_and_row_order`, `test_hybrid_xa_judge_main_with_matching_results`.

**No documentation in `docs/configs/run-config.md` or dataset-config docs** for this new mode. PRs #453 and #459 got the same note — easier to add docs at PR time than to remember later.

**PR description "Verification" is functional/manual only** — no automated test references. Worth at least pointing to a scenario YAML that exercises the hybrid path so others can repro the test. Also: if you confirm the `uv run --isolated` import issue (inline comment on `hybrid_xa_judge.py:10`), worth describing in the PR how the manual verification got past that — I may be misreading the invocation.

evalbench/databases/bigquery.py
```diff
@@ -27,6 +27,9 @@ def __init__(self, db_config):
         self.client = bigquery.Client(project=self.project_id)
         self.tmp_users = []
 
+    def ensure_database_exists(self):
```
IsmailMehdi: **Signature violates the abstract base class.** `databases/db.py:182` declares:
```python
@abstractmethod
def ensure_database_exists(self, database_name: str) -> None:
```
All 8 other implementations (mysql, postgres, sqlite, spanner, bigtable, mongodb, sqlserver) match this signature. The new BQ override drops the parameter entirely. Any generic caller that does `db.ensure_database_exists("foo")` — which is the documented contract — will hit `TypeError: ensure_database_exists() takes 1 positional argument but 2 were given`.

```python
def ensure_database_exists(self, database_name: str) -> None:
    # BigQuery datasets are project-scoped; no per-database creation needed.
    pass
```

evalbench/scorers/sqlite_bridge.py
```diff
@@ -0,0 +1,52 @@
...
+    parent_dir = os.path.dirname(__file__)
+    root_dir = os.path.abspath(os.path.join(parent_dir, "..", ".."))
+    db_dir = os.path.join(root_dir, "db_connections", "bird")
+    if not os.path.exists(db_dir):
+        return []
+
+    candidates = [
+        f[:-7] for f in os.listdir(db_dir) if f.endswith(".sqlite")
+    ]
+    for cand in candidates:
```
IsmailMehdi: **Picks an arbitrary SQLite file by trial-and-error.** Iterating every `.sqlite` file in the directory and using whichever one accepts the query is wrong in two ways:

1. **Non-deterministic.** Returns whatever file `os.listdir` yielded first that didn't raise. If two BIRD databases (`california_schools`, `card_games`) happen to have tables/columns with overlapping names, the resolved "truth" depends on the OS's directory ordering. Run the same scenario on two machines, get two different scores.
2. **Silent wrong-answer risk.** A query may "succeed" against the wrong DB by returning bogus rows that don't match the scenario's intent. The judge then scores against those rows and reports a misleading PASS/FAIL.

The scenario record already carries the database name (the `database` field in BIRD dataset entries — `california_schools`, etc.). Pass it through:
```python
def get_sqlite_ground_truth(query: str, database: str) -> list:
    sqlite_path = os.path.join(db_dir, f"{database}.sqlite")
    if not os.path.exists(sqlite_path):
        return []
    conn = sqlite3.connect(sqlite_path)
    try:
        return pd.read_sql_query(query, conn).to_dict(orient="records")
    finally:
        conn.close()
```
Bonus: dramatically faster — one connect instead of N.

evalbench/scorers/sqlite_bridge.py
```diff
@@ -0,0 +1,52 @@
...
+
+import pandas as pd
+import yaml
+
+
+def get_sqlite_ground_truth(query: str) -> list:
+    """Resolves candidate SQLite database files and executes query."""
+    parent_dir = os.path.dirname(__file__)
+    root_dir = os.path.abspath(os.path.join(parent_dir, "..", ".."))
+    db_dir = os.path.join(root_dir, "db_connections", "bird")
```
IsmailMehdi: **Hardcoded path.** `db_connections/bird` is specific to one dataset and won't work for other suites. The experiment_config already carries `database_configs`; derive the SQLite path from there instead of hardcoding the directory. As written, this scorer is silently bird-only — and nothing in the docstring says so.

evalbench/scorers/sqlite_bridge.py
```diff
@@ -0,0 +1,52 @@
...
+            df_cand = pd.read_sql_query(query, conn)
+            conn.close()
+            return df_cand.to_dict(orient="records")
+        except Exception:
+            continue
+
+    return []
+
+
+def is_hybrid_cross_db_enabled() -> bool:
```
IsmailMehdi: **Sniffing `sys.argv` to detect mode** is fragile and creates several real problems:

- **Only works for `python evalbench.py --experiment_config=foo.yaml`.** Other invocation paths don't have this argv pattern: the gRPC server (`eval_server.py`) constructs configs programmatically, `run_suite` passes config paths via a different mechanism, and any programmatic / in-process caller doesn't have argv at all. Hybrid mode silently disables in these cases.
- **Re-parses the YAML on every call.** `compare()` runs once per scenario per trial; this is O(N) file reads of the same YAML.
- **Cross-scorer coupling via string match.** Checks the *PythonScorer*'s `script_path` to decide what *LLMRater* does. If someone renames `hybrid_xa_judge.py` or moves it, LLMRater silently goes back to returning 0 on golden errors.

The signal that hybrid mode is active is conceptually a scorer config, not a global mode. Lift it to `LLMRater.__init__`:
```python
def __init__(self, config, global_models):
    ...
    self.hybrid_ground_truth = config.get("hybrid_ground_truth", False)
```
And in YAML:
```yaml
scorers:
  llm_rater:
    hybrid_ground_truth: true
```
Then the `compare()` branch checks `self.hybrid_ground_truth`. No argv sniffing, no cross-scorer coupling, no file I/O in the hot path.

evalbench/scorers/llmrater.py
```diff
@@ -252,7 +252,15 @@ def compare(
             return 100, "Skipped. Exact Match was found."
 
         if golden_error:
-            return 0, "Golden query failed to execute."
+            # If using hybrid judge, fetch ground truth from SQLite when BQ
+            # fails on golden query syntax (e.g. SQLite functions in reference
+            # queries).
+            if sqlite_bridge.is_hybrid_cross_db_enabled():
```
IsmailMehdi: **Silent fallback — no log line when it fires.** When hybrid mode kicks in and resolves a golden answer from SQLite, nothing in the logs records that this happened. The eval row just has a normal-looking score. Debugging "why does my BQ run report fewer golden errors than expected" becomes a source-dive.

Add one log line:
```python
if sqlite_bridge.is_hybrid_cross_db_enabled():
    logging.info(
        "Hybrid ground truth: BQ golden query failed, resolving from "
        "SQLite reference. query=%s", golden_query
    )
    golden_execution_result = ...
```

hybrid_xa_judge.py
```diff
@@ -0,0 +1,109 @@
+"""Hybrid Execution Accuracy (XA) Cross-Database Evaluator for EvalBench."""
+
+from decimal import Decimal
+import json
+import sqlite3
+import sys
+
+import pandas as pd
+
+from evalbench.scorers.sqlite_bridge import get_sqlite_ground_truth
```
IsmailMehdi: **This import will fail at runtime.** `pythonscorer.py:57` invokes the script as:
```python
command = ["uv", "run", "--isolated", self.script_path]
```
`--isolated` strips the parent's PYTHONPATH and runs in a clean venv with no access to the `evalbench` source tree. `from evalbench.scorers.sqlite_bridge import ...` will raise `ModuleNotFoundError`. The script returns nonzero, the scorer reports a generic `FAIL: Script failed with exit code...`, and the user never sees a real result.

The PR description mentions functional end-to-end testing, so either I'm misreading the invocation OR the testing didn't actually exercise this code path. Worth confirming with a fresh-venv repro.

Fix: inline the ~10 lines of `get_sqlite_ground_truth` directly into `hybrid_xa_judge.py` so the script has no project-local imports. Together with #2's fix (pass `database` in), the inlined function becomes very small.

evalbench/scorers/judges/hybrid_xa_judge.py
```diff
@@ -0,0 +1,109 @@
+"""Hybrid Execution Accuracy (XA) Cross-Database Evaluator for EvalBench."""
```
IsmailMehdi, line 1: **Script lives at the repo root.** All other Python under this repo lives under `evalbench/` or `datasets/`. Convention break. Suggest moving to something like `evalbench/scorers/judges/hybrid_xa_judge.py` and updating the `script_path` in any example configs that reference it.

evalbench/scorers/judges/hybrid_xa_judge.py
```diff
@@ -0,0 +1,109 @@
...
+    def normalize_df(df: pd.DataFrame) -> list[tuple]:
+        rows = []
+        for _, r in df.iterrows():
+            normalized_row = []
+            for val in r:
+                if pd.isna(val):
+                    normalized_row.append(None)
+                elif isinstance(val, (int, float, Decimal)):
+                    try:
+                        normalized_row.append(round(float(val), 4))
```
IsmailMehdi, line 56: **`compare_result_sets` has three lossy behaviors worth pinning down in the docstring** so the next reader doesn't second-guess them:

1. **Floats rounded to 4 decimals** (`round(float(val), 4)`). For some BIRD queries — AVG/SUM over money or rate fields — the reference may exceed 4 decimal places, turning a real mismatch into a false PASS. Either widen the precision or document the choice.
2. **Sort key `lambda x: str(x)`** sorts `[1, 10, 2]` as `["(1,)", "(10,)", "(2,)"]`. Fine for equality (both sides sort identically) but unreadable in FAIL logs. Consider `key=lambda r: tuple(str(v) for v in r)` for slightly less surprising debug output.
3. **`.0` suffix stripped from string values.** If BQ returns the literal string `"3.0"` (e.g., a version label) and SQLite returns `"3"`, they're considered equal. Almost certainly intentional for numeric-as-string coercion but worth a one-line comment.

A short docstring listing these explicitly makes the contract clear and lets future maintainers reason about edge cases without re-deriving them.

IsmailMehdi:
Maybe this should be built as a tool instead of inline with evals. wdyt ?

evalbench/databases/bigquery.py
arieljassan: Updated the override signature in `BigQuery` to match `databases/db.py` by accepting `database_name: str` to maintain the abstract base class contract and prevent potential `TypeError` crashes.

evalbench/scorers/sqlite_bridge.py
arieljassan: Replaced the search loop with direct database file lookup. The query runner now accepts the specific `database` identifier as an argument and connects directly to `f"{database}.sqlite"` in the target directory.

evalbench/scorers/sqlite_bridge.py
arieljassan: Deleted the `sys.argv` sniffing and global configuration parsing from `sqlite_bridge.py`. The hybrid mode is now explicitly controlled via the `hybrid_ground_truth` scorer config under `llmrater` in the scenario YAML files, loaded cleanly inside the `LLMRater` constructor.

evalbench/scorers/sqlite_bridge.py
arieljassan: Removed the hardcoded path. The SQLite directory is now dynamically loaded from the connection parameters in `database_configs` and passed down to the resolution utility, keeping the scorer generalized for other datasets.

evalbench/scorers/llmrater.py
arieljassan: Added a clear log line: `logging.warning("BQ golden query failed, falling back to SQLite reference database to resolve ground truth for query: %s", golden_query)` which prints whenever the BigQuery execution fails and falls back to SQLite.

hybrid_xa_judge.py
arieljassan: Inlined the query execution utility `get_sqlite_ground_truth` directly inside `hybrid_xa_judge.py` and removed all local `evalbench` package imports. This makes the judge script completely self-contained, allowing it to run successfully inside isolated virtual environments (`uv run --isolated`).

evalbench/scorers/judges/hybrid_xa_judge.py
arieljassan: Relocated `hybrid_xa_judge.py` to its standard path: `evalbench/scorers/judges/hybrid_xa_judge.py` and updated script paths in the configurations.

evalbench/scorers/judges/hybrid_xa_judge.py
arieljassan: Updated the `compare_result_sets` docstring to explicitly list the cell-level normalization rules (rounding floats to 4 decimal places, sorting row tuples lexicographically, stripping trailing `.0` string suffixes, and ignoring headers) to make the comparison contract clear for future maintainers.

arieljassan:
Thank you for the review! addressing your general comments:

**No tests for any of the new code.**

Added 7 unit tests covering both components under `evalbench/test/`:
* **`sqlite_bridge_test.py`**: Verifies dynamic SQLite path loading and fallback error resilience when a database file is missing.
* **`hybrid_xa_judge_test.py`**: Verifies all result set cell normalization rules (float rounding to 4 decimal places, string `.0` suffix stripping, lexicographical sorting, ignoring column names, matching float vs Decimal types) and tests stdin/stdout payload JSON parsing under the sandbox execution.


 **No documentation in docs/configs/run-config.md or dataset-config docs for this new mode.**

Added documentation detailing the new configurations:
* **`run-config.md`**: Documented `hybrid_ground_truth` under `llmrater`, added details for configuring the generic `python_scorer`, and documented the cell normalization rules of the included `hybrid_xa_judge.py` script.
* **`dataset-config.md`**: Added a note explaining how the `golden_sql` dialect should map to the execution engine dialect (e.g. `"bigquery"`) even if the reference query is written in SQLite syntax.
* **Example YAML**: Added `datasets/bird/example_hybrid_run_config.yaml` to provide a clear reference on how to construct a hybrid evaluation configuration.


 **PR description "Verification" is functional/manual only — no automated test references.**

Added unit tests and a functional test:

```bash
.venv_eb/bin/python3 -m unittest evalbench/test/sqlite_bridge_test.py evalbench/test/hybrid_xa_judge_test.py
```
And added a note explaining how the sandbox execution (`uv run --isolated`) is handled via duplicating `get_sqlite_ground_truth` directly inside `hybrid_xa_judge.py` to prevent import issues in isolated environments.

evalbench/databases/bigquery.py
```diff
@@ -27,6 +27,10 @@ def __init__(self, db_config):
         self.client = bigquery.Client(project=self.project_id)
         self.tmp_users = []
 
+    def ensure_database_exists(self, database_name: str) -> None:
```
IsmailMehdi: dead code, see line 183.

IsmailMehdi (changes requested):
LGTM once minor dupe method removed

arieljassan (commented):
Addressed comment of code dupe in bigquery.py

IsmailMehdi (approved):

PR #469, opened 2026-06-29
Feat/mcp readability generator

akangsha7:
Adds `McpToolsGenerator`, the fetch half of the MCP readability check. It
  obtains an endpoint's tools and renders them as man-page markup (via the
  formatter from #463) for the readability scorer.
  
  The system-under-test is pluggable via `tools_source.type`:
  - **http** — live `tools/list` over Streamable HTTP (official `mcp` SDK), no auth
  - **stdio** — launch a local MCP server and speak MCP over its stdio pipes
  - **file** — load a raw `tools/list` JSON dump (offline / deterministic runs)
  
  URL normalization adds scheme + `/mcp` suffix. Registers `"mcp_tools"` in
  `get_generator`.
  
  ## Tests
  Unit tests for all three sources, URL sanitization, error paths, and the
  generator entry point. `22 passed` (with the formatter suite).
  
  ## Note
  Stacked on #463 (the man-page formatter). Until #463 merges, this PR's diff
  includes that commit; it collapses to just the generator once #463 lands.

### DIFF
--- a/evalbench/generators/models/mcp_tools.py
+++ b/evalbench/generators/models/mcp_tools.py
@@ -8,10 +8,9 @@ readability *scorer* then evaluates against the style guide.
 The source is pluggable per endpoint via ``tools_source.type``. The type names
 the *transport*, not the protocol -- ``http`` and ``stdio`` both speak MCP:
   - ``http``: live MCP ``tools/list`` over Streamable HTTP using the official
-    ``mcp`` Python SDK. The fetch URL is ``tools_source.url`` (falling back to
-    the endpoint's ``endpoint_url``). No authentication is configured -- the
-    public endpoints are reached unauthenticated; auth can be added back later
-    if a future endpoint requires it.
+    ``mcp`` Python SDK. The fetch URL is ``tools_source.url``. No authentication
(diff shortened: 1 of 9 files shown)


IsmailMehdi (dismissed):
it would be good to have a look at a run config for this.

akangsha7:
Added sample run config

evalbench/generators/models/mcp_tools.py
```diff
@@ -0,0 +1,227 @@
...
+            raise McpToolsError(f"Unknown tools_source.type '{source_type}'")
+
+        return tools, format_tools_to_man_page(tools)
+
+    # ------------------------------------------------------------------
+    # Source resolution
+    # ------------------------------------------------------------------
+    @staticmethod
+    def _resolve_source(endpoint: dict) -> dict:
+        """The endpoint's ``tools_source`` (or ``{}`` to default to ``http``)."""
```
helloeve, line 89: nit: the default to http seems to be a logic in fetch_tools rather than this helper function

helloeve (approved):

PR #470, opened 2026-06-29
Feat/mcp readability metrics scorer

akangsha7:
Adds `McpToolMetricsScorer`, the deterministic (LLM-free) half of the MCP
  readability check. It computes model-independent size/cost metrics over the
  tools returned by `McpToolsGenerator`:
  
  - **`total_tools`** — number of tools the endpoint exposes
  - **`estimated_tokens`** — rough token footprint, `len(JSON(tool)) / 4` summed
    across tools
  - **`token_budget_used_percent`** — the estimate as a percentage of the
    configured `token_budget` (`None` when no positive budget is set; `0.0` for an
    empty toolset with a budget)
  
  ## Design notes
  - Kept separate from the LLM style judge so the metric logic stays deterministic
    and independently testable.
  - Serializes via `model_dump_json(by_alias=True, exclude_none=True)` so the
    estimate reflects actual wire field names (`inputSchema`) and omits unset
    optionals. Accepts both `mcp.types.Tool` objects and raw `tools/list` dicts.
  
  ## Tests
  `test/mcp_tool_metrics_test.py` — counts, the token estimate (exact for dict
  input), budget math, the no-/zero-budget cases, and config-vs-per-call budget
  precedence. Full MCP suite: `33 passed`.
  
  ## Stacking
  Stacked on the generator PR (`McpToolsGenerator`), which is stacked on #463 (the
  man-page formatter). Until those land, this PR's diff includes their commits;
  review/merge bottom-up: #463 → generator → metrics.

### DIFF
--- a/evalbench/scorers/mcp_tool_metrics.py
+++ b/evalbench/scorers/mcp_tool_metrics.py
@@ -1,0 +1,82 @@
+"""Deterministic metrics scorer for an MCP endpoint's tool listing.
+
+This is the *metrics* half of the MCP readability check. Unlike the LLM style
+judge, it computes purely deterministic, model-independent numbers over the tools
+returned by ``McpToolsGenerator``:
+
+  - ``total_tools``: the number of tools exposed by the endpoint.
+  - ``estimated_tokens``: a rough token footprint of the tool definitions,
+    approximated as ``len(JSON(tool)) / 4`` summed across tools.
+  - ``token_budget_used_percent``: that estimate as a percentage of the
+    configured ``token_budget`` (``None`` when no positive budget is set).
+
+It is kept separate from the LLM judge so the metric logic stays deterministic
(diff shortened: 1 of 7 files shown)


evalbench/scorers/mcp_tool_metrics.py
```diff
@@ -0,0 +1,82 @@
...
+
+from collections.abc import Sequence
+import json
+from typing import Any
+
+# Rough chars-per-token heuristic for estimating a tool's token footprint.
+_CHARS_PER_TOKEN = 4
+
+
+class McpToolMetricsScorer:
```
IsmailMehdi, line 26: This class doesn't follow the traditional scorer signature, I'm also not sure how it's going to be use, can you propose how it would be invoked ?

akangsha7:
Folded into #472 (the combined readability eval PR, stacked on #469). The McpToolMetricsScorer from this branch is included there.

PR #472, opened 2026-07-01
feat(mcp-readability): compliance orchestrator, LLM judge, and metrics scorer

akangsha7:
## Summary

The evaluation half of the MCP-readability work, running on top of the `mcp_tools`
generator from #469. Given a set of MCP endpoints, it fetches each endpoint's
tools (rendered as a man page) and scores them against the Data Cloud MCP tool
style guide, emitting one result row per endpoint through the shared EvalBench
reporters.

- **`McpReadabilityOrchestrator`** (`orchestrator: mcp_readability`) — for each
  endpoint: fetch tools via the generator, gather applicable waivers, run every
  scorer declared in the run config, and merge their columns into a single result
  row. The orchestrator is **scorer-agnostic**: it iterates the registered
  scorers named under the run config's `scorers:` block (via `SCORER_REGISTRY`)
  and invokes each — adding a new scorer (e.g. conformance) needs only a registry
  entry and a config block, no orchestrator changes.
- **`mcp_readability_scoring.py`** — the plug-and-play scorer contract:
  `EndpointContext` + the scorer registry the orchestrator drives.
- **`McpToolMetricsScorer`** (`mcp_tool_metrics`) — deterministic tool count /
  estimated tokens / token-budget usage. Owns its own `token_budget` config.
- **`McpStyleReadabilityScorer`** (`mcp_style_readability`) — LLM judge scoring
  the man page vs the style guide (P0/P1/P2 findings, readability score, waived
  rules). Owns its own `model_config` + `style_guide` config.
- **exceptions helper** — per-endpoint rule waivers, matched by
  product_name / endpoint_type against style-guide `{#tag}` rule ids.
- **`dataset.py` / `evalbench.py` plumbing** — `dataset_config` is optional, and
  orchestrators may emit results without NL2SQL scores (None-guarded reporter
  writes); logs an info message when no dataset path is provided.

Run config uses a plug-and-play `scorers:` block; each key names a registered
scorer and carries that scorer's own config:

```yaml
scorers:
  mcp_tool_metrics:
    token_budget: 200000
  mcp_style_readability:
    model_config: datasets/model_configs/gemini_3.1_pro_model.yaml
    style_guide: datasets/mcp_readability/style_guide.md
```

## Testing

- `pytest evalbench/test/mcp_readability_test.py evalbench/test/mcp_tool_metrics_test.py` — **31 passing**.
- Covers exceptions matching, deterministic metrics, the LLM-judge parsing, and
  an offline end-to-end orchestrator run (file source + stubbed LLM) asserting
  the result schema.
- **Local e2e run** against the sample config exercising all three
  `tools_source` types in one job (file / http / stdio): 3/3 endpoints checked,
  0 errors — `mcp_tool_metrics` 3/3, `mcp_style_readability` 2/3.

### DIFF
--- a/datasets/mcp_readability/run_config.yaml
+++ b/datasets/mcp_readability/run_config.yaml
@@ -12,23 +12,31 @@
 
 orchestrator: mcp_readability
 
+# Plug-and-play scorers. Each key names a registered mcp_readability scorer
+# (SCORER_REGISTRY in the orchestrator) and carries that scorer's own config.
+# The orchestrator runs every scorer declared here against each endpoint and
+# merges their result columns; the shared analyzer aggregates a per-scorer pass
+# rate (comparator = scorer name) for the scores / summary reports. Add a scorer
+# (e.g. conformance) by registering the class and adding a block here.
+scorers:
+  # Deterministic size metrics. Pass = within token budget. `token_budget` feeds
+  # mcp_readability_token_budget_used_percent (endpoints may override per-entry).
+  mcp_tool_metrics:
+    token_budget: 200000
+  # LLM judge vs the style guide. Pass = no P0 findings. Owns its model config
+  # and style-guide path. Run with EVAL_GCP_PROJECT_REGION=global.
+  mcp_style_readability:
+    model_config: datasets/model_configs/gemini_2.5_pro_model.yaml
+    style_guide: datasets/mcp_readability/style_guide.md
+
 # Inputs (paths relative to the repo root / run cwd).
 endpoints_config: datasets/mcp_readability/endpoints.yaml
 exceptions_config: datasets/mcp_readability/exceptions.yaml          # optional
 tools_generator_config: datasets/mcp_readability/tools_generator.yaml
 
-# Token budget for token_budget_used_percent (endpoints may override per-entry).
-token_budget: 200000
-
 # Optional filter: only check endpoints whose endpoint_type is in this list.
 # Empty = check all. e.g. [PROD]
 endpoint_types: []
-
-# LLM judge: consumes a model config and the style-guide path. Reuses the
-# shared repo model config.
-readability_judge:
-  model_config: datasets/model_configs/gemini_2.5_pro_model.yaml
-  style_guide: datasets/mcp_readability/style_guide.md
 
 # Per-endpoint check concurrency.
 runners:
--- a/evalbench/dataset/dataset.py
+++ b/evalbench/dataset/dataset.py
@@ -242,6 +242,17 @@ def load_json(json_file_path):
 
 
 def load_dataset_from_json(json_file_path, config):
+    # No dataset path (e.g. orchestrators driven by their run config rather than a
+    # prompt dataset): nothing to load. flatten_dataset({}) yields []. Log it so
+    # a run that *did* expect a dataset (missing/misconfigured dataset_config)
+    # doesn't fail silently.
+    if not json_file_path:
+        logging.info(
+            "load_dataset_from_json: no dataset path provided; returning an "
+            "empty dataset. Expected only for orchestrators that drive their "
+            "own inputs from the run config."
+        )
+        return {}
     input_items = {}
     dataset_format = config.get("dataset_format", "evalbench-standard-format")
     if dataset_format == "bird-interact-format":
--- a/evalbench/evaluator/mcp_readability/orchestrator.py
+++ b/evalbench/evaluator/mcp_readability/orchestrator.py
@@ -1,0 +1,318 @@
+"""Orchestrator for the MCP style-guide readability check.
+
+Flow per endpoint:
+  1. The tools generator fetches the endpoint's tools and renders the man-page
+     markup.
+  2. Applicable exceptions (waivers) are gathered.
+  3. Every scorer declared under ``scorers:`` in the run config is invoked with
+     the shared per-endpoint context; each contributes result-row columns and a
+     binary summary score (see ``scorers.mcp_readability_scoring``).
+  4. A result row is assembled from the base identity columns plus every
+     scorer's contribution.
+
+The orchestrator is scorer-agnostic: it instantiates whatever scorers the run
+config declares (via ``SCORER_REGISTRY``) and merges their contributions. Adding
+a new scorer (e.g. conformance testing) needs only a registry entry and a
+run-config block -- no changes here.
+
+Failure handling is fail-fast: if any endpoint cannot be fetched or scored, the
+exception propagates and the whole job aborts with nothing persisted. There is no
+per-endpoint status; the failure surfaces in the run log. Isolate flaky or
+experimental endpoints with a separate run config rather than per-row state.
+
+Like the standard EvalBench orchestrators, :meth:`process` dumps the result rows
+to a temp JSON and returns it as ``results_tf``; ``evalbench.py`` then hands that
+to the shared reporters (CSV / BigQuery). The orchestrator performs no direct
+CSV or BigQuery writes itself.
+"""
+
+import concurrent.futures
+import datetime
+import json
+import logging
+import tempfile
+import threading
+
+from evaluator.orchestrator import Orchestrator
+from generators.models import get_generator
+from scorers.mcp_readability_scoring import EndpointContext
+from scorers.mcp_style_readability import McpStyleReadabilityScorer
+from scorers.mcp_tool_metrics import McpToolMetricsScorer
+from util.config import load_yaml_config
+
+from evaluator.mcp_readability import exceptions as exceptions_mod
+
+
+# Registered mcp_readability scorers, keyed by the name used under ``scorers:`` in
+# the run config (also each scorer's ``comparator`` in the summary). Add a new
+# scorer here plus a run-config block to extend the check -- the orchestrator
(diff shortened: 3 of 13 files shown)


evalbench/dataset/dataset.py
```diff
@@ -242,6 +242,10 @@ def load_json(json_file_path):
 
 
 def load_dataset_from_json(json_file_path, config):
+    # No dataset path (e.g. orchestrators driven by their run config rather than a
+    # prompt dataset): nothing to load. flatten_dataset({}) yields [].
+    if not json_file_path:
+        return {}
```
helloeve, line 255: maybe adding a info message here? so it won't silently fail for other use cases where a dataset is indeed expected from the run config

evalbench/evaluator/mcp_readability/enums.py
```diff
@@ -0,0 +1,53 @@
...
+
+They are defined as plain Python enums here so the feature is self-contained;
+values can be aligned to an external schema later without touching call sites.
+All enums are written to the results CSV using their ``.name``.
+"""
+
+from enum import Enum
+
+
+class EndpointType(Enum):
```
helloeve: Do we actually need this enum? Since we controls the logic that parses the run config, we can just simply reject values that are not supported -- or assigning a default (If it turns out there is a typo in the run config, the developer who author that should either realize it from the eval run exception, or realizing the eval result were not categorized right in the dashboard)

evalbench/evaluator/mcp_readability/orchestrator.py
```diff
@@ -0,0 +1,350 @@
...
+        type_filter = config.get("endpoint_types") or []
+        self.endpoint_type_filter = {
+            coerce_endpoint_type(t).name for t in type_filter
+        }
+
+        # Shared model registry for get_generator (lock + cache), as used by
+        # the standard orchestrators.
+        self.global_models = {
+            "lock": threading.Lock(),
+            "registered_models": {},
```
helloeve, line 127: what is this used for?

evalbench/evaluator/mcp_readability/orchestrator.py
```diff
@@ -0,0 +1,350 @@
...
+
+        # Exceptions (waivers).
+        self.all_exceptions = exceptions_mod.load_exceptions(
+            config.get("exceptions_config")
+        )
+
+        # Default token budget + endpoint_type filter. Normalize the filter
+        # values through the same coercion used for endpoints so that qualified
+        # (ENDPOINT_TYPE_PROD), lower-case, or numeric spellings all match.
+        self.default_token_budget = config.get("token_budget", 0) or 0
```
helloeve: nit: this should just be `token_budget` as you are reading it and assign default if missing

evalbench/evaluator/mcp_readability/orchestrator.py
```diff
@@ -0,0 +1,350 @@
...
+        if not tools_generator_config:
+            raise ValueError("mcp_readability requires 'tools_generator_config'")
+        self.tools_generator = get_generator(
+            self.global_models, tools_generator_config
+        )
+
+        # Deterministic metrics scorer (model-independent).
+        self.metrics_scorer = McpToolMetricsScorer(
+            {"token_budget": self.default_token_budget}
+        )
```
helloeve: nit: shall we have `token_budget` part of the scorer config? or maybe under readability_judge? Following this it is slightly better to understand which are the configurable knobs which may impact the final number for the eval task under a single config block

evalbench/evaluator/mcp_readability/orchestrator.py
```diff
@@ -0,0 +1,350 @@
...
+
+        # LLM readability judge. The run config groups its settings under
+        # ``readability_judge`` (model config + style-guide path).
+        judge_config = config.get("readability_judge") or {}
```
helloeve: can this be {}? IIUC, we will at least want the style guide and some model declared for this to work

evalbench/evaluator/mcp_readability/orchestrator.py
```diff
@@ -0,0 +1,350 @@
...
+        self.style_guide = self._read_text(judge_config.get("style_guide"))
+        self.scorer = McpStyleReadabilityScorer(
```
helloeve: Could we make the orchestrator a bit de-coupled from the scorer? Ideally we should have the orchestrator to iterate through all registered scorers from within the run config, and then invoke each to produce some metrics -- rather than orchestrator explicitly initialize dedicated ones.

From the run config, this is what I have in mind based on your current run config:

```
scorers:
  style_compliance:
    token_budget: 2000
  mcp_style_readability:
    model_config: datasets/model_configs/gemini_2.5_pro_model.yaml
    style_guide: datasets/mcp_readability/style_guide.md
```

Following this, you should have a more extensible scoring-logic to follow the plug-and-play approach; The current implementation would requires sizable changes to this orchestrator when adding new scorer (since we will likely to introduce additional scorers like conformance testing, etc.)

evalbench/dataset/dataset.py
akangsha7: Done

evalbench/evaluator/mcp_readability/enums.py
akangsha7: Makes sense. Removed

evalbench/evaluator/mcp_readability/orchestrator.py
akangsha7: It's the get_generator model registry (lock + registered_models cache): get_generator memoizes model clients there so scorers/threads reuse one client instead of re-instantiating per call. Clarified the comment.

evalbench/evaluator/mcp_readability/orchestrator.py
akangsha7: Removed — token_budget now lives under the mcp_tool_metrics scorer's own config block.

evalbench/evaluator/mcp_readability/orchestrator.py
akangsha7: Done

evalbench/evaluator/mcp_readability/orchestrator.py
akangsha7: Done

evalbench/evaluator/mcp_readability/orchestrator.py
akangsha7: Refactored to be scorer-agnostic

evalbench/evaluator/mcp_readability/orchestrator.py
helloeve: sg!

helloeve (dismissed):
Overall LGTM! It would be good to conduct a local run with the sample config and validate things working e2e.

Also seeking another review from @IsmailMehdi as this introduces new orchestration logic to evalbench.

datasets/mcp_readability/run_config.yaml
```diff
@@ -12,24 +12,32 @@
...
+# (e.g. conformance) by registering the class and adding a block here.
+scorers:
+  # Deterministic size metrics. Pass = within token budget. `token_budget` feeds
+  # mcp_readability_token_budget_used_percent (endpoints may override per-entry).
+  mcp_tool_metrics:
+    token_budget: 200000
+  # LLM judge vs the style guide. Pass = no P0 findings. Owns its model config
+  # and style-guide path.
+  mcp_style_readability:
+    model_config: datasets/model_configs/gemini_2.5_pro_model.yaml
```
prernakakkar-google, line 29: nit: prefer using 3.1 pro now

datasets/mcp_readability/run_config.yaml
```diff
@@ -12,24 +12,32 @@
...
+    style_guide: datasets/mcp_readability/style_guide.md
```
prernakakkar-google, line 30: Do we plan to have multiple style guides here or just this one?

prernakakkar-google:
PR description needs to be updated as per changes.

datasets/mcp_readability/run_config.yaml
akangsha7: just one

datasets/mcp_readability/run_config.yaml
akangsha7: Done — switched to `gemini_3.1_pro_model.yaml`.

akangsha7:
Updated the PR description to match the current implementation (scorer-agnostic orchestrator + plug-and-play `scorers:` block, `McpStyleReadabilityScorer`, enums removed) and added the local e2e run results across all three tools_source types (file/http/stdio).

datasets/mcp_readability/run_config.yaml
akangsha7: Moving to  3.1, wold require changes to the judge:

  1. Set a high max_output_tokens in _generate's GenerateContentConfig so verbose findings don't get cut off.
  2. Inspect finish_reason — if MAX_TOKENS, raise a clear "response truncated" error instead of a cryptic JSON parse error.
  
  Leaving this for followup

IsmailMehdi (approved):

helloeve (approved):

prernakakkar-google (approved):

PR #492, opened 2026-07-13
Add field to track time to first response and token usages

totoleon:


erinlimbogoogle (approved):

evalbench/eval_service.py
```diff
@@ -517,4 +519,31 @@ def _process_results(
                 "p90": round(latencies.quantile(0.9), 2),
             }
 
+    # Add percentiles for metrics reported via the `other` map. `other` is
+    # stored as a string-serialized dict in the dataframe, so parse it first.
+    if "other" in results_df.columns:
```
IsmailMehdi, line 524: Any pointer to where this "other" is coming from ?  Should this also be in the standalone evalbench ?

IsmailMehdi (changes requested):

PR #494, opened 2026-07-14
fea(dea): Add support for tools metrics for DEA

minzznguyen:
### Description
Implement logic to extract tool names and details by:
-  decode the final conversation token in `gcp_data_engineering_agent.py`
- add fields to store tools information in `dataengineeringagentinput.py`

### Recorded accumulated_tools when running `example_run_config.yaml`

```python
[
    'provide_recommendations',
    'validate_dataform_action',
    'compile_dataform',
    'write_dataform_file',
    'get_schema_and_samples',
    'assess_cleaning_quality'
]
```

#### Captured Execution Details (tools_by_name)

* **write_dataform_file**: 4 execution calls captured (`definitions/simple_constant_view.sqlx`, `workflow_settings.yaml`, `package.json`)
* **compile_dataform**: 6 execution calls captured (including all parameters and error stack traces)
* **validate_dataform_action**: 1 call captured
* **assess_cleaning_quality**: 1 call captured
* **get_schema_and_samples**: 1 call captured
* **provide_recommendations**: 3 calls captured

### DIFF
--- a/evalbench/evaluator/dataengineeringagentevaluator.py
+++ b/evalbench/evaluator/dataengineeringagentevaluator.py
@@ -219,18 +219,14 @@ class DataEngineeringAgentEvaluator:
                 if fail:
                     tool_data["fail"] = 1
 
-            agent_json = {
-                "response": agent_text,
-                "stats": {
+            conversation_history.append({
+                "user": current_prompt,
+                "agent": agent_text,
+                "agent_stats": {
                     "tools": {
                         "byName": tools_by_name
                     }
-                }
-            }
-
-            conversation_history.append({
-                "user": current_prompt,
-                "agent": json.dumps(agent_json),
+                },
             })
 
             # Simulated User checks conversation plan and generates next prompt
--- a/evalbench/evaluator/dataengineeringagentevaluator.py
+++ b/evalbench/evaluator/dataengineeringagentevaluator.py
@@ -200,9 +200,37 @@ class DataEngineeringAgentEvaluator:
                 agent_text
             )
 
+            this_turn_tools = getattr(
+                eval_result, "this_turn_tool_details", []
+            )
+
+            tools_by_name = {}
+            for t in this_turn_tools:
+                name = t["name"]
+                params = t["params"]
+                output = t.get("output", {})
+                fail = t.get("fail", 0)
+
+                tool_data = tools_by_name.setdefault(
+                    name, {"parameters": [], "outputs": [], "fail": 0}
+                )
+                tool_data["parameters"].append(params)
+                tool_data["outputs"].append(output)
+                if fail:
+                    tool_data["fail"] = 1
+
+            agent_json = {
+                "response": agent_text,
+                "stats": {
+                    "tools": {
+                        "byName": tools_by_name
+                    }
+                }
+            }
+
             conversation_history.append({
                 "user": current_prompt,
-                "agent": agent_text,
+                "agent": json.dumps(agent_json),
             })
 
             # Simulated User checks conversation plan and generates next prompt
@@ -316,7 +344,9 @@ class DataEngineeringAgentEvaluator:
             "prompt": scenario["starting_prompt"],
             "conversation_history": json.dumps(conversation_history, indent=2),
             "scenario": scenario,
-            "accumulated_tools": [],
+            "accumulated_tools": getattr(
+                eval_result, "accumulated_tools", []
+            ),
             "accumulated_skills": [],
             "job_id": job_id,
             "metadata": metadata,
--- a/evalbench/generators/models/gcp_data_engineering_agent.py
+++ b/evalbench/generators/models/gcp_data_engineering_agent.py
@@ -1,9 +1,11 @@
 import asyncio
+import base64
 import collections.abc
 import concurrent.futures
 import datetime
 import json
 import logging
+import re
 import threading
 import uuid
 from typing import Any, Coroutine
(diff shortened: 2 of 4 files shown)


graceqi-g (dismissed):

evalbench/generators/models/gcp_data_engineering_agent.py
```diff
@@ -261,7 +338,12 @@ def generate_internal(self, prompt: EvalDeaRequest) -> EvalDeaRequest:
         )
 
         try:
-            prompt.generated_nl_response = self.run_async(coro)
+            reply_text, new_token = self.run_async(coro)
+            prompt.generated_nl_response = reply_text
+            if new_token:
+                all_tools = _extract_tool_details_from_token(new_token)
```
IsmailMehdi, line 391: **Blocker — the conversation token is cumulative, so this double-counts on turn 2+.**

The whole point of `_conversation_token_cache[conversation_id] = new_conversation_token` (L458) is that the server-side conversation state is preserved across turns via a round-tripped token. The token returned on turn N encodes the history *up to and including* turn N — not just turn N's delta. So `_extract_tool_details_from_token(new_token)` on turn 2 returns every `functionCall` from turns 1 and 2 combined, and `tools_by_name` in the evaluator counts turn 1's calls a second time.

The PR body's numbers (`compile_dataform: 6 calls`, `write_dataform_file: 4 calls`) were captured from a single-turn scenario in `example_run_config.yaml`, which is why this didn't surface. A 2-turn scenario would show it immediately.

Options (in order of preference):
1. Diff by `functionCall.id` against a per-request set of already-seen ids, kept on `EvalDeaRequest`.
2. Move extraction to end-of-conversation in `_finalize_scenario` and skip the per-turn breakdown entirely.
3. Rename `this_turn_tool_details` to something honest (`accumulated_tool_details`) and de-double-count in `process_scenario` before building `tools_by_name`.

Worth confirming the token is cumulative with one real 2-turn run before choosing — if by some chance it *is* per-turn, the field name is right and this comment is moot.

evalbench/generators/models/gcp_data_engineering_agent.py
```diff
@@ -261,7 +338,12 @@ def generate_internal(self, prompt: EvalDeaRequest) -> EvalDeaRequest:
...
+                prompt.this_turn_tool_details = all_tools
+                prompt.accumulated_tools = list({t["name"] for t in all_tools})
```
IsmailMehdi: **Blocker — this doesn't accumulate; it overwrites.**

On turn 2 this replaces `prompt.accumulated_tools` with only turn 2's tool names, dropping everything from turn 1. Given the field name, the `list[str] = []` initialization on `EvalDeaRequest`, and the "accumulated_" semantics used elsewhere in the codebase, the intent is clearly a union across turns.

Fix:
```python
prompt.accumulated_tools = list(dict.fromkeys(
    [*prompt.accumulated_tools, *(t["name"] for t in all_tools)]
))
```

Use `dict.fromkeys`, not `list({...})` — set iteration order isn't guaranteed stable, and downstream snapshot tests / diffs will flake if the order shifts.

(Note: this bug is currently masked by the cumulative-token issue on the extract line above — the token *already* contains all prior tools, so overwriting with the cumulative extract accidentally produces the right union. Fixing one without the other will regress the other. Fix them together.)

evalbench/evaluator/dataengineeringagentevaluator.py
```diff
@@ -200,9 +200,37 @@ def process_scenario(
...
+                    "tools": {
+                        "byName": tools_by_name
+                    }
+                }
+            }
+
             conversation_history.append({
                 "user": current_prompt,
-                "agent": agent_text,
+                "agent": json.dumps(agent_json),
```
IsmailMehdi: **Blocker — shape change on `conversation_history[i]["agent"]` is a downstream break.**

Before this PR: `agent` is the raw agent text (a plain string). After: `agent` is `json.dumps({"response": ..., "stats": {...}})` — a JSON-encoded string wrapping the text plus a structured stats dict.

Any consumer that reads `conversation_history[i]["agent"]` as text now silently gets a JSON blob. Please confirm nothing downstream regresses — in particular:
- Scorers under `evalbench/scorers/` that inspect `conversation_history`
- `SimulatedUser.get_next_response(conversation_plan, conversation_history, agent_text)` — if it looks at anything other than the third arg
- Viewer / reporting code that renders transcripts (BigQuery, CSV, HTML)

Strongly consider *not* changing this contract at all. Keep `agent` as plain text and put the structured metrics on a sibling field:

```python
conversation_history.append({
    "user": current_prompt,
    "agent": agent_text,
    "agent_stats": {"tools": {"byName": tools_by_name}},
})
```

That:
- keeps the transcript human-readable,
- lets scorers consume typed data directly without `json.loads` on a JSON-inside-a-JSON,
- doesn't require updating every existing consumer.

If the JSON envelope really is required by some downstream tool, please call out which one in the PR body and update the affected consumers in the same PR.

IsmailMehdi (commented):
Leaving the three blocker-level items inline. The rest of the review (regex fragility, URL-safe base64, missing `.get()` guards, per-turn reset, and a note on extractor test coverage) is posted separately — not repeated here.

evalbench/generators/models/gcp_data_engineering_agent.py
minzznguyen: confirm that the token tool record is cumulative. Addressed using option 1

evalbench/generators/models/gcp_data_engineering_agent.py
minzznguyen: addressed this blocker with blocker 1 together!

evalbench/evaluator/dataengineeringagentevaluator.py
minzznguyen: Yes, I agree that we should not change the contract, addressed by. There is currently no dependency on contracts that require JSON as of now. Addressed by having your recommended structure!

evalbench/test/gcp_data_engineering_agent_test.py
```diff
@@ -469,10 +469,15 @@ def mock_generate(eval_result):
...
-        {"user": "hello", "agent": "Hi user!"},
+        {
+            "user": "hello",
+            "agent": "Hi user!",
+            "agent_stats": {"tools": {"byName": {}}},
+        },
         {
             "user": "What is target workspace?",
             "agent": "It is test-workspace.",
+            "agent_stats": {"tools": {"byName": {}}},
```
graceqi-g, line 480: can we add a more complex UT? it would be nice to test the tool deduplication logic you added is working correctly

IsmailMehdi (approved):

PR #501, opened 2026-07-17
Feature/default and generated columns

aseering:
I thought I had posted this months ago but it seemingly got lost somewhere, apologies!

This change updates the data ingestion logic.  For Spanner:
* When computing the set of columns to load data into, generated columns are now ignored
* Default columns aren't ignored (because loading data into them is valid but optional), but we reserve a couple special keyword names that schemas can use for default ID columns and the loader won't try to populate them

The underlying problem is as follows:  Spanner requires that all tables contain a PK, and that all PK values be unique.  Some of our datasets do not guarantee uniqueness.  The recommended workaround in this case is to construct a synthetic column where each row is a UUID, or a value from a sequence, or otherwise some synthetic and guaranteed-unique value.  (Then just ignore this column; let Spanner manage and use it itself.)

Some of our schemas do this.  Which is fine.

The issue then is, the data loader tries to populate _all_ columns on the table, and it errors out if it can't do so.  Populating generated columns with data from a .csv file isn't a correct thing to do.  (They're generated and managed by the database; they're not allowed to be provided by the user.)  Populating columns with a DEFAULT clause is permissible, but not required, and not desired for this use case.  This PR adds some logic to enable those use cases.

### DIFF
--- a/evalbench/databases/spanner.py
+++ b/evalbench/databases/spanner.py
@@ -513,6 +513,11 @@ class SpannerDB(DB):
                             break
                 if not info:
                     continue
+                # CSV files are headerless, meaning we map values positionally.
+                # If the CSV has C columns and the DB has S columns (where C < S),
+                # we slice to map values to the first C columns. This allows
+                # database columns with defaults (like surrogate_id PKs) at the end
+                # to be omitted from the insert, letting Spanner generate defaults.
                 csv_col_count = len(rows[0]) if rows else 0
                 columns = info["columns"][:csv_col_count]
                 processed_rows = []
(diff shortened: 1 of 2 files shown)


IsmailMehdi (dismissed):

IsmailMehdi:
PTAL [pycodestyle / pycodestyle (pull_request)](https://github.com/GoogleCloudPlatform/evalbench/actions/runs/29612964046/job/87991630165?pr=501)

aseering:
Fixed the pycodestyle issues.

Could someone with access please do a `/gcbrun`?

prernakakkar-google (approved):

PR #503, opened 2026-07-19
feat(query_data_api): add option to use REST API fallback for unreleased proto fields

g-lynnzee:
Goal: Enable context engineering agent to use API fields without client libraries (i.e. fields for trustedtester) and to connect to non GA endpoints (for developing against lower endpoints).

The client (context engineering agent) will hold the logic on when to determine what to set.


evalbench/generators/models/query_data_api.py
```diff
@@ -67,31 +110,74 @@ def generate_internal(self, prompt: str) -> Dict[str, Any]:
...
+        if not credentials.valid:
+            auth_req = google.auth.transport.requests.Request()
+            credentials.refresh(auth_req)
+
+        url = (
+            f"https://{self.api_endpoint}/v1beta/projects/{self.project_id}"
+            f"/locations/{self.location}:queryData"
+        )
+
+        payload = {
```
prernakakkar-google, line 174: context is sent to the REST endpoint as a raw, unvalidated dict, while generationOptions goes through _to_camel_case_dict. If self.context contains nested snake_case keys (as in the test_generate_internal_rest_fallback_arbitrary_nested_context test), they'll be sent as-is and likely silently ignored/no-op by the server rather than surfacing a clear error. Consider either (a) documenting that callers of the REST fallback are responsible for supplying already-camelCased context, or (b) recursively camelCasing context too for consistency with generationOptions.

prernakakkar-google (dismissed):

helloeve (dismissed):

evalbench/generators/models/query_data_api.py
```diff
@@ -67,31 +122,74 @@ def generate_internal(self, prompt: str) -> Dict[str, Any]:
...
+                f"GDA REST API rate limited / unavailable ({resp.status_code})"
+            )
+
+        resp.raise_for_status()
+        data = resp.json()
+
+        return _format_result(
+            data.get("generatedQuery"),
+            data.get("intentExplanation"),
+            data.get("disambiguationQuestion"),
```
wangauone, line 206: Do we want to add the new debug fields as well?

g-lynnzee:
PTAL, fixed camel casing for rest as suggested

evalbench/generators/models/query_data_api.py
g-lynnzee: No, prefer leaving that for another PR. It's a bigger change - we'll need to do add a client lib for parsing given it is a struct.

helloeve (dismissed):

evalbench/generators/models/query_data_api.py
```diff
@@ -67,31 +122,74 @@ def generate_internal(self, prompt: str) -> Dict[str, Any]:
...
+        if not credentials.valid:
+            auth_req = google.auth.transport.requests.Request()
+            credentials.refresh(auth_req)
+
+        url = (
+            f"https://{self.api_endpoint}/v1beta/projects/{self.project_id}"
+            f"/locations/{self.location}:queryData"
+        )
+
+        payload = {
```
wangauone, line 174: Have we tested this? It seems the payload is missing the `parent` field?

g-lynnzee: I tested this E2E by using the new graph_ids field.

It works because HTTP POST will extract fields like parent from the request path, per AAIP 127: "Fields bound to the URL path template (e.g. {parent=projects/*/locations/*}) are extracted directly from the request URL by the Google API Gateway frontend before the request body is parsed. The gateway automatically populates request.parent on the backend protobuf message."

evalbench/generators/models/query_data_api.py
```diff
@@ -67,31 +122,74 @@ def generate_internal(self, prompt: str) -> Dict[str, Any]:
...
             logger.exception("Unhandled exception during QueryData API call")
             raise
+
+    def _query_data_rest(self, prompt: str) -> Dict[str, Any]:
+        """
+        Fallback REST HTTP execution path for unreleased/private fields
+        missing from PyPI SDK protos.
+        """
+        logger = logging.getLogger(__name__)
+        credentials, _ = google.auth.default(
```
IsmailMehdi: i think this will force a refresh on every call, it makes a new credentials.

IsmailMehdi (changes requested):

evalbench/generators/models/query_data_api.py
g-lynnzee: My understanding from docs is that this fetches the credentials. refresh() actually refreshes.

google.auth.default(): https://googleapis.dev/python/google-auth/1.7.0/reference/google.auth.html#google.auth.default
refresh: https://googleapis.dev/python/google-auth/1.7.0/reference/google.auth.credentials.html

g-lynnzee: But I see your point in caching the credentails object across multiple _query_data_rest calls. Making changes...

IsmailMehdi (approved):

PR #505, opened 2026-07-20
Improve turn scorer - add new scorer for agent steps

prernakakkar-google:


evalbench/scorers/agentsteps.py
```diff
@@ -0,0 +1,119 @@
+"""
+AgentSteps Scorer
```
helloeve, line 2: Is this scorer applicable to all agent harness or just claude?

prernakakkar-google: Its harness agnostic.

IsmailMehdi (approved):

PR #507, opened 2026-07-21
Dataset Quality orchestrator

omkargaikwad23:


evalbench/evaluator/dataset_quality/orchestrator.py
```diff
@@ -0,0 +1,216 @@
+"""Orchestrator for intrinsic dataset-quality grading.
```
prernakakkar-google, line 1: Lets stick to evaluation pattern since this is scorer we should use agent orchestrator add NOOPGenerator as discussed before. This should be scorer.

omkargaikwad23: Sg! I have raised a separate PR for the same. PTAL: https://github.com/GoogleCloudPlatform/evalbench/pull/508

PR #508, opened 2026-07-21
feat: add dataset-quality scoring framework for CUJ datasets

omkargaikwad23:
## Summary

  Adds a **dataset-quality (DQ) scoring** framework that grades a CUJ dataset as a
  whole — so authors get an actionable quality signal on their scenarios and tool coverage.

  The whole dataset is bundled into a single `all_cujs` wrapper scenario and routed
  through a new `noop_agent` generator so it reaches the scoring path without
  invoking a real agent. `DatasetQualityScorer` then runs a set of pluggable
  sub-scorers, rolls their scores up into weighted categories, assigns a letter
  grade, and emits one CSV row per category plus a summary row.

  ### Sub-scorers
  - **trajectory_coverage**, **naming_distribution** — static/structural checks
  - **parameter_coverage** — LLM judge: fraction of the tool param surface CUJs exercise
  - **vague_examples** — LLM judge: flags under-specified prompts
  - **error_recovery** — LLM judge: unhappy-path coverage
  - **composition** — LLM judge: multi-tool + sequencing dependency
  - **cuj_diversity** — LLM judge: spread across the five interaction paths

  Each returns a `SubScoreContribution` (score, applicability, row fields,
  suggestions, per-CUJ evidence, distributions). Non-applicable metrics (empty
  dataset, missing schema, judge failure) drop out of the weighted average rather
  than tanking the grade.

  ## Test plan
  - [x] `.venv/bin/pycodestyle evalbench/ --config=.pycodestyle` passes
  - [x] Run DQ scorer against a sample dataset; verify per-category + summary rows emit
  - [x] Verify judge-failure path yields a non-applicable metric (not a crash / F)
  - [x] Verify synthesis report category scores match the numeric rows
  - [x] Existing generator tests (agy_cli, query_data_api) still pass


prernakakkar-google:
Fix: Run pycodestyle  evalbench --config=.pycodestyle
evalbench/scorers/dataset_quality/llm.py:91:41: E203 whitespace before ':'
1       E203 whitespace before ':'
Error: Process completed with exit code 1.

evalbench/scorers/dataset_quality/trajectory_coverage.py
```diff
@@ -0,0 +1,60 @@
...
+import logging
+
+from scorers.dataset_quality.context import (
+    CATEGORY_TOOL_ACTIVATION,
+    DatasetQualityContext,
+    SubScoreContribution,
+)
+
+
+class TrajectoryCoverageScorer:
```
prernakakkar-google: trajectory_coverage only measures the MCP-tool channel: _fetch_tools() queries setup.mcp_servers only (and logs setup.extensions as unsupported), and it reads only expected_trajectory. Skills (setup.skills/extensions → expected_skills → accumulated_skills, already modeled by skillstrajectorymatcher) never count. The design's Tool Activation Faithfulness category explicitly includes skills. 

Suggestion: make the catalog a union (tools ∪ skills) and add a sibling skill_coverage scorer under CATEGORY_TOOL_ACTIVATION, each marking applicable=False when its catalog is empty. The weighted-average-over-applicable rollup then handles mcp-only / skill-only / both with no special-casing. Note this is static presence-in-trajectory, not the design's "SKILL.md was actually read" / tool-blind-delta checks — worth calling out as a follow-up

evalbench/scorers/dataset_quality/scorer.py
```diff
@@ -0,0 +1,195 @@
...
+                eval_results = json.loads(eval_results) if eval_results else {}
+            except json.JSONDecodeError:
+                eval_results = {}
+        scenario = (eval_results or {}).get("scenario", {}) or {}
+        scenarios = scenario.get("all_cujs") or []
+        if not scenarios:
+            logging.warning("dataset_quality: no CUJs found in wrapper scenario")
+        return scenarios
+
+    def _fetch_tools(self) -> list:
```
prernakakkar-google, line 287: If _fetch_tools() raises (network/ADC blip), it propagates to the outer loop in score.py, which records comparison_error and leaves score 0. For a global grade shown on the trends page, a transient failure reads as a real F. Suggest distinguishing infra failure from a computed grade.

evalbench/scorers/dataset_quality/llm.py
```diff
@@ -0,0 +1,152 @@
...
+
+import json
+import logging
+import re
+
+
+# Max output for the target Gemini Pro models. Gemini 3.x draws reasoning tokens
+# from the output budget, so a high cap keeps thought tokens from starving the
+# JSON answer (which otherwise finishes with empty ``.text``).
+_MAX_OUTPUT_TOKENS = 65536
```
prernakakkar-google, line 12: Each judge gets the whole dataset in one prompt and returns one tag per CUJ, bounded by _MAX_OUTPUT_TOKENS. For hundreds of CUJs this risks truncated output → parse failure → all-default → silently wrong low score. If you find larger datasets use chucking/batching

datasets/dataset_quality/example_run_config.yaml
```diff
@@ -0,0 +1,78 @@
...
+# `gcloud auth application-default login`). Run the LLM judges with
+# EVAL_GCP_PROJECT_REGION=global.
+scorers:
+  dataset_quality:
+    product_name: "Gemini CLI Tools"
+    model_config: datasets/model_configs/gemini_cli_model.yaml
+    # tools_timeout: 30   # optional; live tools/list fetch timeout in seconds
+    # Each sub-scorer's weight defaults in its scorer class; the commented value
+    # below is that default. Uncomment `weight:` to override.
+    scorers:
```
prernakakkar-google: Weights sum to 115; composition (30) + error_recovery (28) ≈ half the total, and composition averages multi-tool% with the rarely-satisfied sequencing% while carrying the highest weight. A typical happy-path dataset lands at F on ~50% of weight.

evalbench/scorers/dataset_quality/scorer.py
omkargaikwad23: Makes sense! I have now added try catch block and if failure occurs we would be emitting one ungraded row.

evalbench/scorers/dataset_quality/llm.py
omkargaikwad23: Noted. We can implement this in followup PR. Thanks for pointing it out!

datasets/dataset_quality/example_run_config.yaml
omkargaikwad23: Reduced the composition weight to 15 & now total weight sum is 100. Weights sum to 115 won't be an issue since we would be dividing by sum of all weights so even if anyone drops any scorer. All score weights would work as expected.

evalbench/scorers/dataset_quality/trajectory_coverage.py
omkargaikwad23: Good catch! The skill_coverage scorer + applicable=False rollup is definitely the right approach.

The blocker is that we lack a static skill catalog for the denominator setup.skills just has install actions, and the DQ path runs a no-op without a live install.

I've added a note to trajectory_coverage explicitly calling out the MCP-only scope. Let's tackle skill coverage as a follow-up once we figure out how to statically enumerate skills. Sound good?

datasets/dataset_quality/example_run_config.yaml
```diff
@@ -0,0 +1,79 @@
...
+#
+# One run grades one dataset. To grade several, list several run configs under a
+# suite (see datasets/sample_suite_config.yaml).
+
+orchestrator: agent
+
+# Bundles every CUJ into one wrapper scenario so the scorer is invoked once with
+# the full dataset (the default formats can't parse CUJ scenarios this way).
+dataset_format: dataset-quality-format
+dataset_config: datasets/gemini-cli-tools/gemini-cli.evalset.json
```
prernakakkar-google, line 15: I was expecting this to be run from eval_run_suite.yaml. Why are we taking specific dataset here

datasets/dataset_quality/example_run_config.yaml
```diff
@@ -0,0 +1,79 @@
...
+############################################################
+# The `dataset_quality` scorer fetches the product's tool catalog live from its own
+# model_config (every setup.mcp_servers entry with an httpUrl is queried via MCP
+# tools/list; names are namespaced <server>__<tool> to match expected_trajectory).
+# google_credentials servers use ADC, so this needs credentials + network (e.g.
+# `gcloud auth application-default login`). Run the LLM judges with
+# EVAL_GCP_PROJECT_REGION=global.
+scorers:
+  dataset_quality:
+    product_name: "Gemini CLI Tools"
```
prernakakkar-google: Why Gemini CLI Tools

datasets/dataset_quality/example_run_config.yaml
```diff
@@ -0,0 +1,79 @@
...
+    model_config: datasets/model_configs/gemini_cli_model.yaml
```
prernakakkar-google, line 36: This should have been model file like gemini 3.1 pro

datasets/dataset_quality/example_run_config.yaml
```diff
@@ -0,0 +1,79 @@
...
+    # tools_timeout: 30   # optional; live tools/list fetch timeout in seconds
+    # Each sub-scorer's weight defaults in its scorer class; the commented value
+    # below is that default. Uncomment `weight:` to override.
+    scorers:
```
prernakakkar-google: We should rename this field as there is scorers.scorers can be confusing for people.

evalbench/generators/models/mcp_tools.py
```diff
@@ -34,6 +34,7 @@
 
 from .generator import QueryGenerator
```
prernakakkar-google, line 35: File level comment: Why are we making changes here. This generator is for MCP readability.

evalbench/scorers/dataset_quality/prompts/composition_coverage.py
```diff
@@ -0,0 +1,79 @@
+COMPOSITION_COVERAGE_PROMPT = """\
+You are an expert evaluator of conversational AI evaluation datasets. You are given
+an ENTIRE dataset of Critical User Journeys (CUJs): each CUJ is one user-agent test
+scenario, which may be single- or multi-turn. For EVERY CUJ, make two independent
```
prernakakkar-google: nit: remove -

evalbench/scorers/dataset_quality/prompts/composition_coverage.py
```diff
@@ -0,0 +1,79 @@
...
+judgments about how it exercises TOOL COMPOSITION.
+
+Why this matters: real tasks rarely resolve with a single isolated tool call. Strong
+datasets test whether the agent can chain tools together, pass data across
+skills/tools, and respect ordering constraints. A dataset of one-tool-per-scenario
+requests overstates how well the product handles realistic, composite work. Your
+judgments drive a composition-coverage score.
+
```
prernakakkar-google, line 18: Do we need this?

evalbench/scorers/dataset_quality/prompts/composition_coverage.py
```diff
@@ -0,0 +1,79 @@
...
+{cujs_json}
+
+### Output Format
+Return ONLY a JSON object (no markdown, no prose) with exactly this shape:
+{{
+  "tags": [
+    {{
+      "id": "<the CUJ id, copied verbatim from the input>",
+      "is_multi_tool": true|false,
+      "has_sequence_dependency": true|false
```
prernakakkar-google: also add reason field

evalbench/scorers/dataset_quality/prompts/composition_coverage.py
```diff
@@ -0,0 +1,79 @@
...
+  rows; check out the branch before editing; fetch an id before using it).
+- false: the tools are independent / order does not matter, or there is only one
+  operation.
+- Note: needing multiple tools (Judgment 1 = true) does NOT by itself imply an
+  ordering constraint. Judge ordering separately.
+
+When genuinely borderline on either judgment, answer false.
+
+### Input Data
+**Available Tools:** {tool_names}
```
prernakakkar-google, line 45: WHy just tool names we should load skills as well

evalbench/scorers/dataset_quality/prompts/cuj_path_classification.py
```diff
@@ -0,0 +1,115 @@
...
+    PATH_ITERATIVE_REFINEMENT,
+    PATH_ERROR_RECOVERY,
+    PATH_OUT_OF_DOMAIN,
+)
+
+
+CUJ_PATH_CLASSIFICATION_PROMPT = """\
+You are an expert evaluator of conversational AI evaluation datasets. You are given
+an ENTIRE dataset of Critical User Journeys (CUJs): each CUJ is one user-agent test
+scenario, which may be single- or multi-turn. Classify EVERY CUJ into EXACTLY ONE of
```
prernakakkar-google: remove -

evalbench/scorers/dataset_quality/prompts/cuj_path_classification.py
```diff
@@ -0,0 +1,115 @@
...
+the five CUJ paths defined below.
+
+Why this matters: most datasets only test the "Happy Path" -- a perfectly phrased
```
prernakakkar-google, line 23: Do we need this

evalbench/scorers/dataset_quality/prompts/cuj_path_classification.py
```diff
@@ -0,0 +1,115 @@
...
+question that yields a perfect answer -- which gives false confidence. A healthy
+dataset also exercises "unhappy paths" and edge cases (ambiguity, refinement,
+error recovery, guardrails). Your classification drives a dataset-diversity score,
+so judge what each scenario ACTUALLY exercises, not what it superficially resembles.
+
+### CUJ PATHS (choose exactly one per CUJ)
+- "Happy": Standard query. A direct, well-formed request that leads straight to an
+  accurate response. Necessary as a baseline. Example: "Who are our top customers?"
```
prernakakkar-google: getting an error is also accurate. I think we should frame it better.

prernakakkar-google (changes requested):

evalbench/scorers/dataset_quality/prompts/cuj_path_classification.py
```diff
@@ -0,0 +1,115 @@
...
+  / what name; or "top customers?" -> "by revenue or by purchase volume?".
+- "Iterative Refinement": The user progressively narrows or adjusts scope across
+  turns, building on the agent's previous outputs. Example: "Show Q1 sales" ->
+  "actually, exclude enterprise accounts" -> "now filter to EMEA".
+- "Error Recovery": The interaction centers on pivoting after something goes wrong
+  -- a query/tool fails, returns an error, hits a nonexistent resource, or the user
+  identifies a mistake and asks the agent to correct course. Example: "That query
+  failed with a syntax error, can you fix the join?".
+- "Out-of-Domain": The user requests something out-of-scope or disallowed that the
+  agent should refuse or redirect -- e.g. PII/sensitive data, disallowed actions, or
```
prernakakkar-google: may not be the case

evalbench/scorers/dataset_quality/prompts/cuj_path_classification.py
```diff
@@ -0,0 +1,115 @@
...
+the request is legitimate but execution fails or the user corrects a mistake and the
+scenario is about recovering (e.g. a valid query references a column that turns out
+not to exist, then the user asks to fix it).
+
+Do not invent behavior that is not implied by the scenario. When genuinely
+uncertain between an edge path and Happy, prefer the edge path only if the
+conversation_plan clearly describes that dynamic; otherwise classify as "Happy".
+
+### Input Data
+**Available Tools:** {tool_names}
```
prernakakkar-google, line 80: Include skills as well

evalbench/scorers/dataset_quality/prompts/cuj_path_classification.py
```diff
@@ -0,0 +1,115 @@
...
+      "cuj_path": "Happy | Ambiguity & Clarification | Iterative Refinement | Error Recovery | Out-of-Domain"
+    }}
+  ]
+}}
+Return one entry in "tags" for EVERY CUJ in the input. Each "id" MUST match an input
+CUJ id exactly. Each "cuj_path" value MUST be exactly one of the five strings listed
+above."""
+
+
+CUJ_PATH_CLASSIFICATION_SCHEMA = {
```
prernakakkar-google, line 119: Here and elsewhere do we want separate I think I would just want 1 result not per CUJ as it wastes output tokens

evalbench/scorers/dataset_quality/prompts/error_recovery_coverage.py
```diff
@@ -0,0 +1,97 @@
...
+MODE_NULL_RESULTS = "Null Results"
+MODE_INCOMPLETE_DATA = "Incomplete Data Set"
+MODE_CASCADING_FAILURE = "Cascading Failure"
+ERROR_RECOVERY_MODES = (
+    MODE_ACCESS_DENIED,
+    MODE_LATENCY_THROTTLING,
+    MODE_OUTPUT_VALIDATION,
+    MODE_NULL_RESULTS,
+    MODE_INCOMPLETE_DATA,
+    MODE_CASCADING_FAILURE,
```
prernakakkar-google: I think we cannot categorize if we want to keep it general

evalbench/scorers/dataset_quality/scorer.py
```diff
@@ -0,0 +1,252 @@
+"""Bridge scorer that grades a CUJ dataset's intrinsic quality.
```
prernakakkar-google, line 1: if _extract_cujs returns [] (missing/malformed all_cujs, or filtered to empty), every scorer returns applicable=False, overall is None, and the summary row is emitted as 0 → an F on the trends page — the exact confusion the tool-discovery branch avoids. Suggest: mirror the tool-discovery treatment — emit score=None + graded: False when context.n == 0, instead of 0.

evalbench/scorers/dataset_quality/prompts/parameter_coverage.py
```diff
@@ -0,0 +1,82 @@
+PARAMETER_COVERAGE_PROMPT = """\
+You are an expert evaluator of conversational AI evaluation datasets. You are given
+the full TOOL SCHEMA for a product and an ENTIRE dataset of Critical User Journeys
+(CUJs): each CUJ is one user-agent test scenario, which may be single- or
```
prernakakkar-google: nit: remove

evalbench/scorers/dataset_quality/prompts/parameter_coverage.py
```diff
@@ -0,0 +1,82 @@
...
+
+### Output Format
+Return ONLY a JSON object (no markdown, no prose) with exactly this shape:
+{{
+  "coverage": [
+    {{
+      "tool": "<tool name, copied verbatim from the schema>",
+      "parameter": "<parameter name, copied verbatim from the schema>",
+      "cuj_ids": ["<id of each CUJ that exercises this parameter>"]
+    }}
```
prernakakkar-google, line 70: add reason of score here and elsewhere

evalbench/scorers/dataset_quality/prompts/parameter_coverage.py
```diff
@@ -0,0 +1,82 @@
...
+silently. A named parameter that appears in zero (or very few) CUJs is a coverage
+gap. Your judgments drive a parameter-coverage score, so ground every decision in
+what each scenario actually demands.
+
+### What "exercises a parameter" means
+A CUJ exercises a tool's parameter when the scenario would require the agent to
+supply a meaningful, scenario-specific value for that parameter while calling that
+tool -- i.e. the request cannot succeed as described unless that parameter is set.
+- Count a parameter for a CUJ only if that parameter's tool is actually used by the
+  CUJ. Use expected_trajectory as the authoritative signal for which tools a CUJ
```
prernakakkar-google: Why?

evalbench/scorers/dataset_quality/prompts/parameter_coverage.py
```diff
@@ -0,0 +1,82 @@
...
+  CUJ under every parameter it genuinely exercises.
+
+When genuinely borderline on whether an optional parameter is exercised, do NOT count
+it.
+
+### Input Data
+**Tool Schema (JSON):** A list of tools. Each tool has "name" and an "inputSchema"
+whose "properties" are the named parameters and whose "required" lists the mandatory
+ones.
+{tool_schema}
```
prernakakkar-google, line 41: add for skills as well

evalbench/scorers/dataset_quality/prompts/vague_examples.py
```diff
@@ -0,0 +1,67 @@
...
+
+### How to judge
+Base your judgment primarily on the starting_prompt (how the user first phrases the
+task), then the conversation_plan. A CUJ is VAGUE if the user's initial goal is
+non-obvious -- expressed by intent rather than by naming the tool/operation/target.
+If the request explicitly names a tool from the available tools, or spells out the
+exact operation and target, it is DIRECT. When genuinely borderline, prefer DIRECT.
+
+### Input Data
+**Available Tools:** {tool_names}
```
prernakakkar-google, line 36: add skills

evalbench/scorers/dataset_quality/composition.py
```diff
@@ -0,0 +1,102 @@
...
+        for s in context.scenarios:
+            sid = s.get("id")
+            tag = tags.get(sid, {})
+            if tag.get("is_multi_tool") is True:
+                multi_ids.append(sid)
+            if tag.get("has_sequence_dependency") is True:
+                seq_ids.append(sid)
+        n_multi, n_seq = len(multi_ids), len(seq_ids)
+
+        multitool_score = round(n_multi / n * 100, 2)
```
prernakakkar-google: multitool_score is the raw share of multi-tool CUJs; a product that exposes one tool cannot have any, so composition scores ~0 and drags the weighted grade — even though the dataset is fine. Every other schema-derived scorer guards this (parameter_coverage → applicable=False when no params; trajectory → when no tools). Suggest: applicable=False when len(context.tools) < 2

prernakakkar-google (changes requested):

datasets/dataset_quality/example_run_config.yaml
omkargaikwad23: That logic will be present in internal repo. This is for open source users if they want to evaluate their dataset quality locally.

datasets/dataset_quality/example_run_config.yaml
omkargaikwad23: Good catch! I will update it to Cloud SQL. This is just example run config referring gemini-cli.evalset.json dataset file.

datasets/dataset_quality/example_run_config.yaml
prernakakkar-google: We are evaluating for internal use case even then it should not be one json.

datasets/dataset_quality/example_run_config.yaml
omkargaikwad23: Agree! I will rename inner block from scorers: to sub_scorers:

datasets/dataset_quality/example_run_config.yaml
omkargaikwad23: This file is required to get the mcp_servers block details & skills path / repo URL.

evalbench/scorers/dataset_quality/prompts/composition_coverage.py
omkargaikwad23: Happy to add it, but to avoid cluttering the card I'd keep it out of the rendered evidence. Either log-only for debugging or a BQ-only column for audit. Which use case are you looking for to include reason field?

evalbench/scorers/dataset_quality/prompts/composition_coverage.py
omkargaikwad23: I will work on skills related change in another PR. I am focusing on making this framework ready including UI changes.
Already added a reply [ref](https://github.com/GoogleCloudPlatform/evalbench/pull/508#discussion_r3642962056).

evalbench/scorers/dataset_quality/scorer.py
omkargaikwad23: Good catch! Fixed by emitting score=None + graded: False before tool discovery so it doesn't post a 0/F.

evalbench/generators/models/mcp_tools.py
omkargaikwad23: The dataset_quality scorer needs the product's live tool catalog for its coverage checks, and this class  already owns the MCP tools/list fetch (SDK, auth, transports). Rather than duplicate that, I reused it — the 
new fetch_tools_from_mcp_servers just aggregates + namespaces (<server>__<tool>) across the configured servers. So it's shared MCP-fetch plumbing now, not readability-only.

evalbench/scorers/dataset_quality/prompts/cuj_path_classification.py
omkargaikwad23: Sure! will work on skill scorer in separate CL.

datasets/dataset_quality/example_run_config.yaml
omkargaikwad23: We would aggregating json files in internal repo and passing combined json.

evalbench/scorers/dataset_quality/prompts/cuj_path_classification.py
omkargaikwad23: Yes, giving LLM judge the purpose of a metric helps it resolve borderline cases towards metric's intent rather than pattern-matching wording.

evalbench/scorers/dataset_quality/prompts/parameter_coverage.py
omkargaikwad23: Remove '-' after single word

evalbench/scorers/dataset_quality/prompts/parameter_coverage.py
omkargaikwad23: Will work on it in another CL. I have some doubts to clarify regarding skills.

evalbench/scorers/dataset_quality/prompts/parameter_coverage.py
omkargaikwad23: Reworded to explain it: whether a tool fires is a fact to read off expected_trajectory, not something to infer from the prompt prose (which can mention things that sound like a parameter without that tool ever running). Anchoring to the trajectory prevents crediting coverage for tools that never run. Also merged the duplicate positive/negative restatement into one line.

evalbench/scorers/dataset_quality/prompts/cuj_path_classification.py
omkargaikwad23: Good catch! Reframed Happy around what really defines it: a straightforward single-pass success with no ambiguity, refinement, error, or guardrail dynamic.

evalbench/scorers/dataset_quality/prompts/cuj_path_classification.py
omkargaikwad23: Reframed Out-of-Domain to be relative to the product's intended scope rather than naming PII as always-disallowed, and kept it only as a scope dependent eg.

evalbench/scorers/dataset_quality/prompts/composition_coverage.py
omkargaikwad23: I have trimmed it. The line "Strong datasets test whether the agent can chain tools together… and respect ordering constraints" was restating both judgments before the JUDGMENT sections define them, so I dropped it.

Kept the rest: all five dataset-quality prompts use the same three-part rationale block (why the dimension matters → what a weak dataset looks like → judge what the scenario ACTUALLY requires) and this one is now the same length as its siblings. Happy to revisit the pattern across all five if you'd rather we drop it everywhere

evalbench/scorers/dataset_quality/composition.py
omkargaikwad23: Agreed on the concern, fixed.

evalbench/scorers/dataset_quality/prompts/cuj_path_classification.py
omkargaikwad23: Makes sense! I will invert the output to label -> ids instead of row per CUJ.

evalbench/scorers/dataset_quality/prompts/error_recovery_coverage.py
omkargaikwad23: Good catch — the old names were data-warehouse flavored. Renamed them to describe how the call fails rather than what comes back, so they apply to any tool-calling product:

-     Latency / Throttling → Transient Failure (also covers timeouts and temporary unavailability)
-     Output Validation Failure → Malformed Output
-     Null Results → Empty Result
-     Incomplete Data Set → Partial Result

  Access Denied and Cascading Failure are unchanged, already generic.

  What this doesn't fix: the denominator is still a hardcoded six, so a product that
  structurally can't produce partial or transient failures is capped at 4/6 = 67% on our
  heaviest-weighted sub-scorer — that's a property of the product, not of the dataset.
 
Deriving the mode list per product from the tool catalog can considered in future scope!

evalbench/scorers/dataset_quality/prompts/parameter_coverage.py
omkargaikwad23: The judge never sees the score: it only emits raw tags, and the scorer computes
  covered/total deterministically. So the field can honestly be "why I tagged it this
  way", not "why the score is X". The score-level narrative is already what the
  synthesis pass produces from the full report, so I want to avoid building a second
  explanation layer that says the same thing.

datasets/dataset_quality/example_run_config.yaml
prernakakkar-google: We should read it from eval_run_suite by using run_config as we cannot do it for all

evalbench/generators/models/mcp_tools.py
```diff
@@ -115,23 +116,104 @@ def _from_http(self, source: dict) -> list[mcp_types.Tool]:
...
+        async with streamablehttp_client(
+            url, headers=headers, timeout=self.timeout
+        ) as streams:
             reader, writer, _ = streams
             async with ClientSession(reader, writer) as session:
                 await session.initialize()
                 tools_response = await session.list_tools()
         return list(tools_response.tools)
 
+    @staticmethod
```
prernakakkar-google: move these to agent generator

evalbench/scorers/dataset_quality/prompts/composition_coverage.py
prernakakkar-google: We should explain the reasoning of scores. The number itself does not explain anything to user.

evalbench/scorers/dataset_quality/prompts/composition_coverage.py
```diff
@@ -0,0 +1,70 @@
...
+
+**CUJs (JSON list):** Each object has "id", "starting_prompt", "conversation_plan",
+and "expected_trajectory".
+{cujs_json}
+
+### Output Format
+Return ONLY a JSON object (no markdown, no prose) with exactly this shape -- one key
+per judgment, listing only the ids of the CUJs it is TRUE for:
+{{
```
prernakakkar-google, line 54: Why we need to do separately for each CUJ

evalbench/scorers/dataset_quality/prompts/error_recovery_coverage.py
```diff
@@ -0,0 +1,88 @@
+# Canonical failure/recovery modes, named for how the call fails rather than for
+# any one product's domain, so the taxonomy applies to any tool-calling product.
+# The judge MUST return only these exact strings; the scorer keys coverage off the
+# named constants so they can't desync.
+MODE_ACCESS_DENIED = "Access Denied"
```
prernakakkar-google: Does it cover all the scenarios

datasets/dataset_quality/example_run_config.yaml
omkargaikwad23: This is just for example purpose for open source community! Internally we would be doing as you mentioned (sent CL).

evalbench/scorers/dataset_quality/prompts/composition_coverage.py
omkargaikwad23: Shared screenshots on chat of the UI. We will be explaining each score & how to fix those issues.

datasets/dataset_quality/example_run_config.yaml
omkargaikwad23: Sure! Doing the same in internal repo as we discussed. This is just for example purpose for open source.

evalbench/scorers/dataset_quality/prompts/composition_coverage.py
omkargaikwad23: The output isn't per-CUJ rows anymore -> it's one grouped result for the whole dataset. No field 
names or labels repeated per CUJ, and only positives are listed, so a dataset with few composite CUJs emits almost nothing.
 
We keep the ids rather than a count for two reasons: **LLMs count unreliably, so
the list is how we get a trustworthy number** 

the ids only live in the report JSON and nothing renders them, so the cost buys nothing visible. I'll surface them in the gap/suggestion text and add one reason string per judgment (dataset-level,
not per CUJ) so the score explains itself - which also covers your  reason-field comment above.

evalbench/scorers/dataset_quality/prompts/error_recovery_coverage.py
omkargaikwad23: Trimmed the error modes and kept it generic so it would cover all the scenarios accurately. Also made sure to keep those generic.

prernakakkar-google (approved):

PR #512, opened 2026-07-22
refactor(agy_cli): parse stream-json events instead of transcript logs

omkargaikwad23:
## Summary
   tool calls, status, and conversation id per invocation.
  - Parse the stream even on a non-zero exit: a timed-out/errored run still emits a full stream ending in an ERROR `result` with usage + tool calls, which
   was previously dropped as unparseable raw JSONL.
  - Add optional `print_timeout` config key → agy's `--print-timeout` (Go-duration; default 5m0s) to bound long print-mode waits.
  - Pin the CLI log via `--log-file` to a deterministic path so model detection reads the exact run's log instead of racing on the newest file (removes
  `_latest_log_path`).
  - Update `docs/agy_cli_agent_testing.md` for the stream-json switch and the new `print_timeout` key.

  ## Test plan
  - [x] `pytest evalbench/test/agy_cli_test.py` (46 passing)
  - [x] `pycodestyle --config=.pycodestyle` clean on changed files
  - [x] End-to-end `./evalbench/run.sh` run, including a timeout turn, retains token/tool stats


evalbench/generators/models/agy_cli.py
```diff
@@ -1070,97 +984,84 @@ def _build_tools_stats(self, calls: list, result_for: dict) -> dict:
...
             self.model
             or self._detect_model_from_log()
             or self._DEFAULT_MODEL_LABEL
         )
+        tokens = self._map_usage_tokens(usage)
         return {
             model_name: {
                 "api": {
                     "totalRequests": 1,
                     "totalErrors": 0,
```
prernakakkar-google: propagate an error result into api.totalErrors. The new path explicitly parses streams from failed/timed-out invocations, but _build_models_stats() always emits "totalErrors": 0. Consequently, a run ending with result.status == "ERROR" will retain token and tool stats—as intended—while misleading downstream error-rate reporting as a successful model request. Pass the final result status into _build_models_stats() and set totalErrors to 1 for non-success statuses (and mirror that value in the roles.main bucket if role-level errors are tracked).

evalbench/generators/models/agy_cli.py
```diff
@@ -1070,97 +984,84 @@ def _build_tools_stats(self, calls: list, result_for: dict) -> dict:
...
         """
         model_name = (
             self.model
             or self._detect_model_from_log()
             or self._DEFAULT_MODEL_LABEL
         )
+        tokens = self._map_usage_tokens(usage)
         return {
             model_name: {
                 "api": {
```
prernakakkar-google, line 1036: Should api be there

evalbench/generators/models/agy_cli.py
```diff
@@ -1070,97 +984,84 @@ def _build_tools_stats(self, calls: list, result_for: dict) -> dict:
...
                     "totalRequests": 1,
```
prernakakkar-google, line 1037: why we are fixing to 1.

evalbench/generators/models/agy_cli.py
```diff
@@ -905,152 +883,88 @@ def _detect_model_from_log(self):
...
-         turn_start_ts, turn_end_ts) = self._pair_calls_and_results(turn_steps)
-
-        final_obj["response"] = "\n".join(response_text_parts).strip() \
-            or fallback_response
-
-        # Approximate end-to-end latency from transcript timestamps.
-        total_duration_ms = self._ms_between(turn_start_ts, turn_end_ts)
+        result = next(
+            (e.get("result", {}) for e in events
+             if e.get("event") == self._EVENT_RESULT),
```
prernakakkar-google: validate parsed event types before calling .get(). The parser only catches JSONDecodeError; a valid JSON scalar or array (for example, a CLI warning serialized as "warning" or []) is appended to events. Later, e.get(...) raises AttributeError. _run_agy_cli() catches that at the outer level, leaves result.stdout as raw NDJSON, and parse_response() consequently returns {}—discarding valid result usage and tool events that appeared before/after the malformed line.

omkargaikwad23: Good catch — you're right. json.loads succeeds on a bare scalar/array ("warning", [], 42), so the line was getting appended to events and the first .get() in _parse_stream_json/_collect_tool_calls raised AttributeError, which the outer except  in _run_agy_cli swallowed — dropping the whole parse and leaving raw NDJSON that parse_response then returned as {}.

Fixed by guarding at the source: only dict events are appended, so non-object lines 
are skipped and all downstream consumers stay safe. Added a regression test 
 (test_parse_stream_json_skips_non_object_lines) that interleaves a 
 string/array/number between valid events and asserts the result + tool stats still  parse. All tests pass.

evalbench/generators/models/agy_cli.py
omkargaikwad23: Yes, we also have it in gemini_cli, codex_cli & claude_code as well.

The end_to_end_latency scorer reads latency from  stats.models.<m>.api.totalLatencyMs (endtoendlatency.py:76), so removing the api block would make that scorer report 0.

evalbench/generators/models/agy_cli.py
omkargaikwad23: It's fixed to 1 because agy's stream-json provides aggregate token totals rather than a per-request count, so each turn is treated as one logical request. This matches the behavior of other harnesses (claude_code, codex_cli, gemini_cli). It's included strictly for schema parity since no scorer currently reads totalRequests (they only use api.totalLatencyMs).

evalbench/generators/models/agy_cli.py
omkargaikwad23: Good catch! However, all three sibling harnesses hardcode totalErrors: 0 (claude_code, codex_cli, gemini_cli). So agy currently matches convention. For now, i have added the fix as you have suggested. 
I will create a followup PR to fix in others.

omkargaikwad23 (commented):
Addressed the comments. PTAL.

docs/agy_cli_agent_testing.md
```diff
@@ -182,6 +182,7 @@ reporting:
 |-----|----------|-------------|
 | `generator` | Yes | Must be `agy_cli` |
 | `model` | Optional | agy UI model label, e.g. `"Gemini 3.1 Pro (High)"`. Passed via the `--model` flag. Must be a valid label, not an API id. |
+| `print_timeout` | Optional | Go-duration string (e.g. `"10m0s"`) passed via agy's `--print-timeout`. Omit to use agy's 5m0s default. |
```
prernakakkar-google: Consider validating print_timeout at config-load time, or at least raising a targeted error when the CLI rejects it. Currently arbitrary strings pass straight through, which shifts a configuration mistake to a runtime agy failure

evalbench/generators/models/agy_cli.py
prernakakkar-google: The parser now correctly discards non-object lines, but it still assumes that event["result"] and event["step_update"] are dictionaries. A valid JSON event such as {"event":"result","result":[]} or {"event":"step_update","step_update":"warning"} will reach .get() at lines 921 or 946 and raise AttributeError. _run_agy_cli() catches that broadly, leaving the entire raw NDJSON stream in stdout; then parse_response() returns {}, which discards otherwise-valid usage and tool events. Since this is an external CLI protocol boundary—and the PR already explicitly handles malformed stream lines—the nested payloads should be type-checked too.

docs/agy_cli_agent_testing.md
omkargaikwad23: Agreed. This was unscoped. agy already defaults to 5m0s and nothing overrides it, so I've reverted print_timeout completely rather than patch the validation. Will bring it back with a validate-and-fallback if a real need shows up.

evalbench/generators/models/agy_cli.py
omkargaikwad23: Right. 
Fixed both:
  - result selector now requires isinstance(e.get("result"), dict), so {"event":"result","result":[]} is skipped rather than returning [] past the {} default.
  - _collect_tool_calls now checks isinstance(step, dict) instead of truthiness, so a string step_update is skipped.

omkargaikwad23 (commented):
Thanks for the feedback! Resolved the comments. PTAL.

prernakakkar-google (approved):

PR #518, opened 2026-07-22
Fix GCP credential issues in Codex CLI generator for local MCP tools

silviojr:
## Problem
When running MCP tools that access Google Cloud Platform (GCP) (e.g., BigQuery via `npx`) through the `codex_cli` generator, authentication fails. 

### Root Cause
EvalBench isolates generator executions by overriding the `HOME` environment variable to a "fake home" directory (e.g., `.venv/fake_home_codex`). Standard Google Cloud SDKs look for Application Default Credentials (ADC) at `~/.config/gcloud/application_default_credentials.json`. Since `~` resolves to the fake home, and `codex_cli` was not seeding this directory with credentials, local stdio MCP servers spawned by Codex could not find the ADC file and failed to authenticate.

## Solution
This PR refactors the credential setup logic from `gemini_cli.py` and `agy_cli.py` into `CodexCliGenerator.__init__` to bring it to parity with other generators.

### what is needed to setup logic:
1. **Detect ADC**: Checks for existing ADC either via the `GOOGLE_APPLICATION_CREDENTIALS` environment variable or at the standard path in the user's real home directory.
2. **Mirror Credentials**: Copies the ADC file to `fake_home/.config/gcloud/application_default_credentials.json` inside the sandbox.
3. **Set Config Path**: Sets `CLOUDSDK_CONFIG` to point to the real `gcloud` config directory (if not already set) to ensure `gcloud` CLI commands work seamlessly.


evalbench/generators/models/agent_cli.py
```diff
@@ -15,6 +17,33 @@ class AgentCliGenerator(QueryGenerator):
...
+        if adc_path and os.path.exists(adc_path):
+            fake_gcloud_dir = os.path.join(fake_home, ".config", "gcloud")
+            os.makedirs(fake_gcloud_dir, exist_ok=True)
+            fake_adc_path = os.path.join(
+                fake_gcloud_dir, "application_default_credentials.json"
+            )
+            if os.path.abspath(adc_path) != os.path.abspath(fake_adc_path):
+                shutil.copy2(adc_path, fake_adc_path)
+
+        if "CLOUDSDK_CONFIG" not in env:
```
prernakakkar-google, line 42: claude_code.py:135 points CLOUDSDK_CONFIG at the fake gcloud dir, while this shared 
helper uses the real one. Is that difference intentional?

prernakakkar-google (approved):

evalbench/generators/models/agent_cli.py
silviojr: I noticed that discrepancy and I was afraid of touching Claude for now. 

However, after digging a little bit, I figured out that this shared helper (and therefore gemini_cli, codex etc) may have a bug. Let me explain and let me know if you agree with me. 

1) My intention was to fix the authentication errors raised by MCP servers. MCP servers look for the credentials in `~/.config/gcloud`. In this case, the **fake home**. However the credentials were not there. What I do here is copy them to the fak ehome.  

2) Although `GOOGLE_APPLICATION_CREDENTIALS` is copied to the fake home, notice that both `env["GOOGLE_APPLICATION_CREDENTIALS"]` and `env["CLOUDSDK_CONFIG"]` continue pointing to the "real home". So they were never updated. **However, in claude, we point those env vars to the fake home**

I suppose that claude is doing the right thing , maintainingg the strict isolation that we should see when multiple runs are happening concurrently. 

I can fix that, but I'd like to hear from you first.

prernakakkar-google: You're right that there's an isolation asymmetry between this helper and claude_code, but I don't think we should fix it by stripping down to ADC-only. We do want gcloud itself to work in these runs, not just ADC-based API auth — codex already depends on this today via _fetch_gcloud_access_token, which shells out to gcloud auth application-default print-access-token against the real HOME/gcloud config for MCP endpoints. So the real gcloud config genuinely needs to be reachable.

 The reason claude_code can point both GOOGLE_APPLICATION_CREDENTIALS and CLOUDSDK_CONFIG at the fake home is that it copies the entire gcloud config dir into the fake home (copytree, minus logs/wal) — active_config, configurations/access_tokens.db, credentials.db, plus ADC. That gives it a complete, working config in an isolated location. This helper only copies the single ADC file, so it points CLOUDSDK_CONFIG at the real dir on purpose — a fake dir with just ADC would break gcloud (no active config, no project/account resolution).

So it's not really a bug, it's the partial version of what claude already does correctly. The right end state is to
 bring this shared helper to parity with claude_code: copytree the whole gcloud config into the fake home, then point both env vars at the fake home. That gets us both working gcloud AND the strict isolation you're asking about for concurrent runs.

 I don't think we need to expand this CL to do that, though — it fixes the reported MCP breakage and keeps gemini/agy behavior identical, so no regression. Let's land it as-is and I'll file a follow-up to unify all four generators on the copytree approach (and retire the divergence with claude). Leaving claude untouched here. Sound good?

silviojr: Sounds good! thanks.

PR #519, opened 2026-07-22
Support timeout configuration for Antigravity (agy) CLI generator

silviojr:
**Overview**
The `agy` CLI defaults to a 5-minute timeout for agent execution. Some complex tasks require more time, and previously this behavior would cause the agent to terminate prematurely during evaluation runs. This PR introduces the ability to configure an explicit timeout limit for `agy` execution directly from the model configuration.

**Changes**
*   **`AgyCliGenerator`**: Added parsing for the `timeout` key in `querygenerator_config`.
*   **Command Execution**: Updated `_base_agy_command` to accept a `timeout` argument. If provided, the generator now automatically translates this into the `--print-timeout <timeout>` flag when invoking `agy`.
*   **Documentation**: Updated `docs/agy_cli_agent_testing.md` to document the new `timeout` field in the Model Configuration. Also updated the class docstring in `agy_cli.py`.
*   **Testing**: Added `test_run_command_argv_shape_with_timeout` to `agy_cli_test.py` to verify the argv shaping correctly includes `--print-timeout`. 

**Example Usage**
In an agy model configuration file (e.g., `agy_cli_model.yaml`):
```yaml
generator: agy_cli
model: "Gemini 3.1 Pro"
timeout: "20m"       
```

### DIFF
--- a/docs/agy_cli_agent_testing.md
+++ b/docs/agy_cli_agent_testing.md
@@ -182,6 +182,7 @@ reporting:
 |-----|----------|-------------|
 | `generator` | Yes | Must be `agy_cli` |
 | `model` | Optional | agy UI model label, e.g. `"Gemini 3.1 Pro (High)"`. Passed via the `--model` flag. Must be a valid label, not an API id. |
+| `timeout` | Optional | Timeout duration string, e.g. `"20m"`. Passed via the `--print-timeout` flag. If omitted, defaults to agy's internal default (5 minutes). |
 | `env` | Optional | Environment variables passed to the CLI process |
 | `setup` | Optional | Tool setup block containing `mcp_servers`, `skills`, or `fake_mcp_servers` |
 
--- a/evalbench/generators/models/agy_cli.py
+++ b/evalbench/generators/models/agy_cli.py
@@ -33,7 +33,8 @@ class AgyCliGenerator(AgentCliGenerator):
     """Generator that queries via the Antigravity CLI (``agy``).
 
     The eval turn runs ``agy -p <prompt> --dangerously-skip-permissions
-    --output-format stream-json [--model <label>] [--continue]``. The on-disk
+    --output-format stream-json [--model <label>] [--print-timeout <timeout>]
+    [--continue]``. The on-disk
     layout lives under ``~/.gemini/antigravity-cli/`` (the binary calls this
     ``appDataDir``). Skills are delivered via plugins (see _setup_skills).
     ``--output-format stream-json`` emits newline-delimited events (an
@@ -59,6 +60,7 @@ class AgyCliGenerator(AgentCliGenerator):
         # flag (None -> flag omitted). See _base_agy_command for the value
         # format and resolution semantics.
         self.model = querygenerator_config.get("model")
+        self.timeout = querygenerator_config.get("timeout")
 
         # Order is load-bearing: paths/dirs must exist before the binary
         # installs and settings/auth write into them, and self.env must carry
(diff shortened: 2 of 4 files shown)


docs/agy_cli_agent_testing.md
```diff
@@ -182,6 +182,7 @@ reporting:
 |-----|----------|-------------|
 | `generator` | Yes | Must be `agy_cli` |
 | `model` | Optional | agy UI model label, e.g. `"Gemini 3.1 Pro (High)"`. Passed via the `--model` flag. Must be a valid label, not an API id. |
+| `timeout` | Optional | Timeout duration string, e.g. `"20m"`. Passed via the `--print-timeout` flag. If omitted, defaults to agy's internal default (5 minutes). |
```
omkargaikwad23, line 185: How this will be handled if incorrect value is passed in timeout parameter? The execution should not fail or it should fail early or default value should be used which is 5 mins.
Also, could you please update the example run config file with this new parameter: https://github.com/GoogleCloudPlatform/evalbench/blob/main/datasets/agy-cli-tools/example_run_config.yaml

evalbench/generators/models/agy_cli.py
```diff
@@ -755,6 +759,7 @@ def _merged_env(self, extra: dict | None = None) -> dict:
     def _base_agy_command(
         cli: str, prompt: str, resume: bool = False, model: str = None,
         output_format: str = None, log_file: str = None,
+        timeout: str = None,
```
prernakakkar-google, line 779: nit: could you add a line to the _base_agy_command docstring describing timeout (maps to --print-timeout, e.g. "20m"), matching how model/output_format/log_file are documented right above?

evalbench/generators/models/agy_cli.py
```diff
@@ -779,6 +784,8 @@ def _base_agy_command(
             command += ["--output-format", output_format]
         if log_file:
             command += ["--log-file", log_file]
+        if timeout:
+            command += ["--print-timeout", timeout]
```
prernakakkar-google, line 808: Consider coercing (str(timeout)) or validating the type in __init__, so a timeout: 20 (int) config fails early instead of at exec time

evalbench/generators/models/agy_cli.py
```diff
@@ -441,7 +443,9 @@ def _verify_mcp_runtime(self, configured_servers: list):
         before = set(os.listdir(log_dir)) if os.path.isdir(log_dir) else set()
 
         env = self._merged_env()
-        cmd = self._base_agy_command(self.agy_bin, "ping", model=self.model)
+        cmd = self._base_agy_command(
+            self.agy_bin, "ping", model=self.model, timeout=self.timeout
+        )
```
prernakakkar-google, line 465: Just confirming: applying the configured timeout to the setup ping probe as well as the eval turn is intentional? Makes sense as an upper bound, just flagging the behavior change.

prernakakkar-google (dismissed):

evalbench/generators/models/agy_cli.py
silviojr: I checked the lines below and there's a hardcoded timeout in `subproces.run()` . I think passing this new timeout doesn't make sense here, specially if it's just a probe. I will remove it.

evalbench/generators/models/agy_cli.py
silviojr: done

evalbench/generators/models/agy_cli.py
silviojr: validating the type in init.

docs/agy_cli_agent_testing.md
silviojr: done

prernakakkar-google (approved):

PR #520, opened 2026-07-22
Add allow_extra_skills config option to SkillsTrajectoryMatcher

saurabh-net:
### Summary of Changes

- **Added `allow_extra_skills` Config Flag**: Added `allow_extra_skills` configuration option to `SkillsTrajectoryMatcher` (defaults to `False` for backward compatibility).
- **Skill Coverage Matching**: When `allow_extra_skills: true` is configured, evaluates skill recall/coverage ($\frac{|\text{expected} \cap \text{actual}|}{|\text{expected}|}$) without penalizing extra activated skills.
- **Empty `expected_skills` Handling**: Returns `100.0%` score if `expected_skills` is empty/omitted and `allow_extra_skills: true`, preserving legacy Jaccard behavior when `allow_extra_skills: false`.
- **Unit Tests**: Added unit test suite in `evalbench/test/skillstrajectorymatcher_test.py` covering coverage matching, backward compatibility, and empty expected skill sets.

### DIFF
--- a/evalbench/scorers/skillstrajectorymatcher.py
+++ b/evalbench/scorers/skillstrajectorymatcher.py
@@ -1,8 +1,9 @@
 """
 SkillsTrajectoryMatcher
 
-Compares expected activated skills vs actual activated skill names,
-using the same Jaccard/Levenshtein approach as TrajectoryMatcher.
+Compares expected activated skills vs actual activated skill names.
+Supports flexible coverage matching (when ignore_extra_skills is True),
+Jaccard set similarity (default), and strict Levenshtein sequence alignment.
 """
 from typing import Tuple, Any, List
 from scorers import comparator
@@ -22,6 +23,7 @@ class SkillsTrajectoryMatcher(comparator.Comparator):
         self.name = "skills_trajectory"
         self.config = config
         self.enforce_order = config.get("enforce_order", False)
+        self.ignore_extra_skills = config.get("ignore_extra_skills", False)
 
     def _levenshtein_distance(self, seq1: List[str], seq2: List[str]) -> int:
         n, m = len(seq1), len(seq2)
@@ -85,8 +87,8 @@ class SkillsTrajectoryMatcher(comparator.Comparator):
         if not isinstance(expected, list) or not isinstance(actual, list):
             return 0.0, "Skills data must be lists."
 
-        if not expected and not actual:
-            return 100.0, "Both expected and actual skill lists are empty."
+        if not expected:
+            return 100.0, "No expected skills specified."
 
         if self.enforce_order:
             distance = self._levenshtein_distance(expected, actual)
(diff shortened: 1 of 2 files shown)


evalbench/scorers/skillstrajectorymatcher.py
```diff
@@ -85,8 +87,8 @@ def compare(
         if not isinstance(expected, list) or not isinstance(actual, list):
             return 0.0, "Skills data must be lists."
 
-        if not expected and not actual:
-            return 100.0, "Both expected and actual skill lists are empty."
+        if not expected:
```
graceqi-g: should we make the condition `if not expected and not actual and not ignore_extra_skills` instead?
i think the updated if statement as is is changing the existing behavior of the scorer

evalbench/scorers/skillstrajectorymatcher.py
```diff
@@ -22,6 +23,7 @@ def __init__(self, config: dict):
         self.name = "skills_trajectory"
         self.config = config
         self.enforce_order = config.get("enforce_order", False)
+        self.ignore_extra_skills = config.get("ignore_extra_skills", False)
```
graceqi-g: nit: can we rename the param to something like allow_extra_skills? it makes reading boolean statements easier, e.g. right now `not ignore_extra_skills` is like a double negative, whereas i find `not allow_extra_skills` a bit easier to understand

graceqi-g (changes requested):

evalbench/scorers/skillstrajectorymatcher.py
saurabh-net: Makes sense, thanks!

evalbench/scorers/skillstrajectorymatcher.py
saurabh-net: Right, I've broken it down to the earlier conditions.

graceqi-g (dismissed):

evalbench/scorers/skillstrajectorymatcher.py
```diff
@@ -99,6 +109,16 @@ def compare(
                 f"(Distance: {distance}, Max: {max_len}). "
                 f"Expected: {expected}, Actual: {actual}"
             )
+        elif self.allow_extra_skills:
```
prernakakkar-google, line 112: Please either (a) document the precedence explicitly (enforce_order takes priority), or (b) reject the invalid combo in __init__ with a clear error, or (c) support ordered+extra.

evalbench/scorers/skillstrajectorymatcher.py
```diff
@@ -1,8 +1,9 @@
 """
 SkillsTrajectoryMatcher
 
-Compares expected activated skills vs actual activated skill names,
-using the same Jaccard/Levenshtein approach as TrajectoryMatcher.
+Compares expected activated skills vs actual activated skill names.
```
prernakakkar-google, line 4: Can you also add corresponding run config to test this out

prernakakkar-google (dismissed):

evalbench/scorers/skillstrajectorymatcher.py
prernakakkar-google: You can ignore this I saw the cl.

evalbench/scorers/skillstrajectorymatcher.py
saurabh-net: Done

prernakakkar-google (approved):

PR #531, opened 2026-07-28
feat: add Dataset Quality tab with automated precompute caching & UI support

omkargaikwad23:
## Summary
  Surfaces the dataset-quality grades produced by the DQ scorer in the viewer.

  - **New "Dataset Quality" tab** : fleet summary (products graded, total CUJs) plus a table of the                  latest report — summary, CUJ path distribution, and a card
    per category with sub-scores, gaps and recommendations.
  - **Detail routing** : a DQ run has no agent trajectory, so
    selecting one renders the report instead of the empty
    Dashboard/Configs/Conversations panes.
  - **Precompute** (`viewer/precompute_dataset_quality.py`, `run_precompute.py`):
    the table view is served from a cache refreshed on the existing precompute
    loop; each precompute is now guarded separately so one failing doesn't
    starve the other. Gaps/evidence/recommendations stay out of the cache and
    are re-read from the run's `scores.csv` only on the detail view.
  - **Scorer** (`evalbench/scorers/dataset_quality/scorer.py`): cast the overall
    and per-category scores to `float` at emit time so the shared FLOAT64
    scores column never receives an int.
  
  ## Test plan
  - [x] Run the precompute and confirm the tab lists a grade per product.
  - [x] Open a product and confirm the report renders with categories, gaps and
        recommendations.
  - [x] Open a non-DQ run and confirm the normal eval tabs still render.
  - [x] Re-run the DQ scorer and confirm scores land in BigQuery without a
        dtype mismatch.

[run-pytest](https://github.com/GoogleCloudPlatform/evalbench/actions/workflows/pytest.yml)


viewer/precompute_dataset_quality.py
```diff
@@ -0,0 +1,181 @@
...
+    values = {}
+    run_time = ""
+    with open(configs_file, newline="") as f:
+        for row in csv.DictReader(f):
+            values[row.get("config", "")] = row.get("value", "")
+            run_time = row.get("run_time", "") or run_time
+    return values, run_time
+
+
+def process_directory(job_id, results_dir):
```
prernakakkar-google, line 74: process_directory returns None when configs.csv/scores.csv don't exist yet (run still in flight), but processed.add(job_id) runs unconditionally afterward. Since the loop only scans d not in processed, any run directory that exists at scan time but hasn't finished writing scores.csv is marked done and never re-examined — its grade never appears in the tab. With a 300s precompute interval and DQ runs that take minutes, this will happen routinely.

omkargaikwad23: Good catch, fixed.

prernakakkar-google (approved):

PR #535, opened 2026-07-28
feat(evalbench): add native Agent Runtime generator support and deployment guide

arieljassan:
### Summary
This PR adds native support for evaluating SQL-generating agents deployed to **Agent Runtime (Gemini Enterprise Agent Platform)** inside the `evalbench` pipeline. It allows developers to deploy their ADK agents and run full evaluations (correctness and execution scoring) directly against the remote live endpoints.
### Key changes
1. **Core Generator**:
   - Implemented `AgentRuntimeGenerator` (`evalbench/generators/models/agent_runtime.py`) which manages client authentication and queries remote Reasoning Engine endpoints using the AI Platform SDK.
   - Registered the `"agent_runtime"` model key in the generator factory (`evalbench/generators/models/__init__.py`).
2. **Pipeline Safety Fix**:
   - Updated `evalbench/work/sqlexecwork.py` to use safe dict retrieval (`self.experiment_config.get("dialect")` instead of `self.experiment_config["dialect"]`). This prevents `KeyError` runner crashes on configs using list-based dialects (such as BIRD SQLite configs).
3. **Deployment guide and example**:
   - Created `docs/examples/agent_runtime_deployment.ipynb` — a step-by-step Colab-friendly notebook explaining how to define, deploy, test, and evaluate a minimal ADK agent against the BIRD dataset.
4. **Configuration templates**:
   - Created default configurations `datasets/model_configs/agent_runtime.yaml` and `datasets/bird/test_agent_runtime_run_config.yaml` to run pipeline evaluations.
5. **Testing**:
   - Added unit tests `evalbench/test/agent_runtime_test.py` covering client authorization setup, remote request payloads, parameters parsing, and response envelope extraction (raw SQL strings vs. JSON envelopes).


evalbench/generators/models/agent_runtime.py
```diff
@@ -0,0 +1,149 @@
...
+        try:
+            data = json.loads(json_match.group(1).strip())
+            if isinstance(data, dict) and "sql" in data:
+                return str(data["sql"])
+        except json.JSONDecodeError:
+            pass
+
+    # Fallback: Split by markdown blocks and take the first segment.
+    if "```" in text:
+        candidate = text.split("```")[0].strip()
```
IsmailMehdi: **Blocking:** this fallback takes the text *before* the first fence, so it never returns the contents of a ```` ```sql ```` block.

This matters because `docs/examples/agent_runtime_deployment.ipynb` explicitly instructs the deployed agent to do exactly that:

> "Wrap your final generated SQL query inside a ```sql code block."

I ran `_extract_sql` against realistic outputs on this branch:

| Input | Returned |
|---|---|
| ```` ```sql\nSELECT COUNT(*) FROM schools;\n``` ```` | `'```sql\nSELECT COUNT(*) FROM schools;\n```'` (fences included) |
| `Here is the query:\n```sql\nSELECT ...\n```` ` | `'Here is the query:'` |
| `Sure!\n```sql\nSELECT 1;\n```\nHope that helps.` | `'Sure!'` |
| ```` ```json\n{"sql": "SELECT 1;"}\n``` ```` | `'SELECT 1;'` :white_check_mark: |

So the documented happy path yields either markdown fences or prose preamble — never SQL. Only the JSON-envelope branch works.

The heuristic is inverted: it should extract content *inside* the fence, not before it. Note the repo already has `util/sanitizer.sanitize_sql` for this — `gemini.py:14` imports it. Reusing that would be preferable to hand-rolling.

The five extraction tests all pass because none of them exercise a ```` ```sql ```` block; worth adding cases with and without a prose preamble.

datasets/model_configs/agent_runtime.yaml
```diff
@@ -0,0 +1,14 @@
+# Agent Runtime (Gemini Enterprise Agent Platform) Generator Manifest
+# Enables native NL2SQL evaluation benchmarking against deployed Agent Runtime endpoints.
+generator: agent_runtime
+
+# Explicit resource URI string (e.g. "projects/12345/locations/us-central1/reasoningEngines/67890")
+# Leave empty string "" to dynamically read MULTI_AGENT_ENGINE_RESOURCE or AGENT_ENGINE_RESOURCE from env.
+resource_name: "projects/743100398377/locations/us-central1/reasoningEngines/7921856483556327424"
```
IsmailMehdi: **Blocking:** this template hardcodes a live Agent Engine resource (project number `743100398377`, engine ID `7921856483556327424`) and a personal argolis project on line 9. `datasets/bird/test_agent_runtime_run_config.yaml:40` hardcodes `arielj-argolis-1` as well.

As shipped, a new user's first run silently targets your sandbox rather than their own project.

The repo convention is env interpolation — see `datasets/model_configs/gcp_data_engineering_agent_model.yaml:2`:

```yaml
gcp_project_id: !ENV ${EVAL_GCP_PROJECT_ID}
```

`pyaml_env` is already a dependency, so this works out of the box. Suggest a placeholder for `resource_name` (the notebook already tells users to paste theirs in at Step 5) and `!ENV` for the project/region.

Related: the comment on line 6 says the resource can be read from `MULTI_AGENT_ENGINE_RESOURCE` or `AGENT_ENGINE_RESOURCE`, but `agent_runtime.py:52` only reads `AGENT_ENGINE_RESOURCE`.

IsmailMehdi (commented):
Nice addition — the generator is clean and factoring the stream parsing into standalone helpers makes it genuinely testable. I ran the new tests locally (`make proto` + CI `PYTHONPATH`) and all 8 pass.

Flagging three things I think block a merge. Details inline.

1. `_extract_sql` doesn't handle ```` ```sql ```` fenced blocks — which is the exact format the accompanying notebook instructs the agent to emit.
2. The committed run config points at a dataset file and a scorer script that don't exist in the repo.
3. The config templates hardcode a personal sandbox project and a live Agent Engine resource ID.

I have a few non-blocking notes too (env var names diverging from `util.gcp`, the blanket `except Exception` defeating `rate_limit` retries, `google-cloud-aiplatform` not being a declared dependency, and the same `["dialect"]` KeyError still present at `interactsqlexecwork.py:113`) — happy to write those up separately if useful.

evalbench/generators/models/agent_runtime.py
arieljassan: Thanks for catching this. I've re-implemented `_extract_sql` to use regex to correctly extract the query inside markdown code fences (handling both plain ```` ``` ```` and ```` ```sql ```` blocks) and then clean it up using the shared `util.sanitizer.sanitize_sql` function.
I've also updated `agent_runtime_test.py` to add new test cases covering raw SQL, simple fenced SQL blocks, and fenced SQL blocks with prose preambles/post-ambles to prevent regressions.

datasets/model_configs/agent_runtime.yaml
arieljassan: You're right. I've updated both `agent_runtime.yaml` and `test_agent_runtime_run_config.yaml` templates to use `!ENV` environment variable interpolation for `resource_name`, `gcp_project_id`, and `gcp_region` (defaulting to standard env vars like `EVAL_GCP_PROJECT_ID` and `AGENT_ENGINE_RESOURCE`).
I also cleaned up the configuration comments to remove references to `MULTI_AGENT_ENGINE_RESOURCE`, which is redundant.

arieljassan:
Thanks for your thorough review and the non-blocking comments. I've addressed all of them in the latest commit:
- **GCP Env Vars**: Updated `agent_runtime.py` to use `get_gcp_project` and `get_gcp_region` from `util.gcp` directly, aligning the environment variables with the rest of the codebase.
- **Rate Limiting**: Replaced the blanket `except Exception` in `generate_internal` by catching `ResourceExhausted` specifically and raising `ResourceExhaustedError` so that the caller's retry/backoff wrapper behaves correctly.
- **Dependencies**: Added `google-cloud-aiplatform` to the dependency list in `pyproject.toml`.
- **KeyError**: Safeguarded the dialect check in `interactsqlexecwork.py:113` using `.get("dialect")` to align with the fix in `sqlexecwork.py` and prevent KeyErrors.

docs/examples/agent_runtime_deployment.ipynb
```diff
@@ -0,0 +1,464 @@
...
+      "execution_count": null,
+      "id": "8adb6260",
+      "metadata": {
+        "id": "8adb6260"
+      },
+      "outputs": [],
+      "source": [
+        "import os\n",
+        "import google.cloud.aiplatform as aiplatform\n",
+        "import agentplatform\n",
```
IsmailMehdi: `agentplatform` is never installed. Step 1 installs `google-cloud-aiplatform google-adk google-auth==2.49.0`, and I confirmed `import agentplatform` raises `ModuleNotFoundError` in a venv with `google-cloud-aiplatform` 1.138 + `google-adk` 1.28. This cell and the `agentplatform.agent_engines.templates.adk` import below it will both fail.

Worth noting the notebook and the generator are on two different SDK surfaces: `AgentRuntimeGenerator` uses `vertexai.agent_engines`, while the notebook uses `agentplatform`. Either add the package that provides `agentplatform` to the Step 1 install, or move the notebook onto `vertexai.agent_engines` so the guide and the generator match.

docs/examples/agent_runtime_deployment.ipynb
```diff
@@ -0,0 +1,464 @@
...
+        "query_types:\n",
+        " - dql\n",
+        "dataset_format: bird-standard-format\n",
+        "\n",
+        "model_config: {temp_model_config_path}\n",
+        "prompt_generator: 'SQLGenBasePromptGenerator'\n",
+        "\n",
+        "scorers:\n",
+        "  python_scorer:\n",
+        "    script_path: '{os.path.join(base_path, 'evalbench/scorers/judges/sqlite_exact_match.py')}'\n",
```
IsmailMehdi: `evalbench/scorers/judges/sqlite_exact_match.py` does not exist. `evalbench/scorers/judges/` contains only `hybrid_xa_judge.py`, and there is no `sqlite_exact_match` anywhere in the tree (the closest match is `evalbench/scorers/exact_match_consistency_comparator.py`, which is a different scorer with a different name).

So the run config generated here points at a missing script and Step 7 fails — which breaks the end-to-end evaluation this notebook exists to demonstrate. Either point at an existing judge or add the scorer in this PR.

IsmailMehdi (commented):
Two blockers in the deployment notebook — both make the documented end-to-end flow fail as written. Details inline.

docs/examples/agent_runtime_deployment.ipynb
arieljassan: I've updated the notebook to use the standard public `vertexai` native SDK APIs (from the standard `google-cloud-aiplatform` library) instead of the private `agentplatform` imports, so it aligns with the generator code. I've also added `pyaml_env` to the setup cell and exported the necessary GCP environment variables so the Colab execution works out-of-the-box. Thanks.

docs/examples/agent_runtime_deployment.ipynb
arieljassan: Thanks for pointing this out. Rather than adding a new custom Python scorer, I updated the notebook's generated run config in step 6 to use the built-in `set_match` and `exact_match` scorers, which work natively for SQLite. I also enabled csv reporting in the temporary config so that the evaluation outputs are written to disk as CSV files. Finally, I updated the model's instructions in `minimal_agent.py` to generate SQLite SQL instead of BigQuery SQL.

IsmailMehdi (approved):

PR #541, opened 2026-08-01
chore: refresh eval configs and correct model-selection docs

omkargaikwad23:
## Summary
  - Bump simulated-user and LLM-scorer models from Gemini 2.5 Pro to 3.1 Pro across the agy-cli datasets
  - Re-enable `token_consumption` for agy runs — stream-json exposes real usage now, so it no longer reports 0
  - Fix stale docs/comments: `--model` accepts both the slug (`gemini-3.1-pro-high`) and the UI label; an unrecognized value hard-fails rather
   than silently falling back; drop obsolete `v1.0.5` pins
- clarify agy authentication and credential mirroring procedures in the testing guide

  ## Test plan
  - [x] Model-selection behavior verified against agy 1.1.9 (both spellings resolve; unrecognized value exits 1)
  - [x] Run `datasets/agy-cli-tools/example_run_config.yaml` end to end

  Docs-and-config only — no behavior change in agy_cli.py beyond docstrings.

### DIFF
--- a/docs/agy_cli_agent_testing.md
+++ b/docs/agy_cli_agent_testing.md
@@ -6,10 +6,9 @@ of [`gemini_cli_agent_testing.md`](gemini_cli_agent_testing.md) and only calls
 out where the two harnesses differ.
 
 > [!IMPORTANT]
-> **First-run auth:** agy uses an OAuth consumer flow backed by the system
-> keyring. Before evals can run, complete `agy`
-> login interactively at least once on the host so a refreshable token
-> exists. After that, the harness can run non-interactively.
+> **First-run auth:** agy has no headless login. Run `agy` once interactively
+> on the host to mint a refreshable token; the harness mirrors it into every
+> eval sandbox after that. See [Authentication](#authentication).
 
 ---
 
@@ -23,6 +22,7 @@ out where the two harnesses differ.
   - [Run Configuration](#1-run-configuration)
   - [Model Configuration](#2-model-configuration)
   - [Evaluation Dataset (Evalset)](#3-evaluation-dataset-evalset)
+- [Authentication](#authentication)
 - [Tool Paradigms](#tool-paradigms)
   - [MCP Servers](#mcp-servers)
   - [Skills](#skills)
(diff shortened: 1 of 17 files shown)


docs/agy_cli_agent_testing.md
```diff
@@ -210,6 +218,45 @@ identical baseline.
...
+> mirrors the **file**, so if the keyring holds your token there is nothing to
+> copy and the run hangs on the login prompt.
+
+### GCP project
+
+agy reads the project for its model backend from `settings.json` -> `gcp.project`,
+**not** from the environment. The harness writes that block into the sandbox from
+`env.GOOGLE_CLOUD_PROJECT` / `env.GOOGLE_CLOUD_LOCATION` in your model config --
+without them every turn returns an empty response and makes no tool calls.
+
```
prernakakkar-google: Should be concise and explainatory for user

omkargaikwad23: Good catch! docs/agy_cli_agent_testing.md:196 already covers it, actionably and without internals details.
Hence removing this block. Thanks!

docs/agy_cli_agent_testing.md
```diff
@@ -210,6 +207,37 @@ identical baseline.
...
+- **Google OAuth** -- personal Google accounts.
+- **Use a Google Cloud project** -- required for managed Workspace or corporate
+  domain accounts, which the plain OAuth path rejects with an auth error.
+
+There is no `agy login` subcommand, service-account flag, or API-key env var for
+agy's own backend. The login writes `antigravity-oauth-token` and
+`installation_id` to `~/.gemini/antigravity-cli/`. The harness copies both into
+the sandbox before every run, so the sandboxed CLI never re-prompts and always
+has a fresh token; the copy is chmod'd to `0600` (owner-only). Only the token is
+load-bearing.
```
prernakakkar-google: Should we have this for external user, I dont think we should put lot of technical details.

omkargaikwad23: Makes sense! Thanks for the review. Addressed in the latest commit.

prernakakkar-google (approved):

PR #549, opened 2026-08-05
authenticate from ADC, drop host OAuth token mirroring

omkargaikwad23:
## Summary

Switches `agy` authentication to Application Default Credentials (ADC) via `AGY_ADC_AUTH` and drops host OAuth token mirroring. It also adds GKE service account key support, missing `quota_project_id` injection, unrecognized model validation, and environment variable stringification.

## Key Changes

- **ADC Authentication**: Updated `_setup_auth` to rely on `AGY_ADC_AUTH` and removed host OAuth token mirroring.
- **GKE Support**: Configured `/etc/evalbench-sa-key/key.json` as a fallback ADC path for GKE environments.
- **Quota Project Injection**: Added `_ensure_quota_project` to automatically inject `quota_project_id` into credentials if missing.
- **Model Validation & Config**: Added `_MODEL_FATAL_MARKER` to detect rejected model selections early during MCP probes and updated the model config to `"Gemini 3.1 Pro (Low)"`.
- **Subprocess Env Fix**: Ensured environment variables are stringified in `_merged_env` to avoid type errors with unquoted YAML scalars.
- **Docs & Tests**: Updated documentation for ADC setup and added unit tests covering ADC staging, GKE keys, quota project injection, and model errors. Cleaned up into a self-contained, standalone guide without repetitive comparison boilerplate.

## Test Plan

- [x] Run `pytest evalbench/test/agy_cli_test.py` to verify unit test coverage for ADC auth and GKE key handling.
- [x] Verified local docker run + GKE run


evalbench/generators/models/agy_cli.py
```diff
@@ -19,6 +19,10 @@
 # location and skips the download when the binary already exists at the target.
```
prernakakkar-google: The current changes will not work with GKE. Please test and fix them.

helloeve (dismissed):

prernakakkar-google (changes requested):

prernakakkar-google:
Description says we cannot use higher model: ```
Re-pin the shipped model to Gemini 3.1 Pro (Low): ADC is entitled to a
smaller model set than an interactive login, and (High) is not in it.
```
is this correct?

**Comment by omkargaikwad23** [comment]
> Description says we cannot use higher model: ``` Re-pin the shipped model to Gemini 3.1 Pro (Low): ADC is entitled to a smaller model set than an interactive login, and (High) is not in it.
> 
> ```
> is this correct?
> ```

Yes, I tested using Gemini 3.1 Pro (High) and it's not supported. I have added a feature for early termination when provided model is incorrect.

**Comment by prernakakkar-google** [comment]
> > Description says we cannot use higher model: ``` Re-pin the shipped model to Gemini 3.1 Pro (Low): ADC is entitled to a smaller model set than an interactive login, and (High) is not in it.
> > ```
> > is this correct?
> > ```
> 
> Yes, I tested using Gemini 3.1 Pro (High) and it's not supported. I have added a feature for early termination when provided model is incorrect.

Is there a way to test it because this is main model to test.

File: evalbench/generators/models/agy_cli.py (hunk 1 above)
Comment (omkargaikwad23) [reply evalbench/generators/models/agy_cli.py]: Thanks for the feedback! Resolved in the latest commit & updated PR description.

**Comment by omkargaikwad23** [comment]
> > > Description says we cannot use higher model: ``` Re-pin the shipped model to Gemini 3.1 Pro (Low): ADC is entitled to a smaller model set than an interactive login, and (High) is not in it.
> > > ```
> > > is this correct?
> > > ```
> > 
> > 
> > Yes, I tested using Gemini 3.1 Pro (High) and it's not supported. I have added a feature for early termination when provided model is incorrect.
> 
> Is there a way to test it because this is main model to test.

We may need to reach out to Agy Team to support those models. I tried with slugs and labels with high effort in model but it's failing.

File: docs/agy_cli_agent_testing.md [hunk 2]
```diff
@@ -209,26 +186,16 @@ identical baseline.
...
-**OAuth token** (its model backend, every turn) and **gcloud ADC** (outbound
-calls from `authProviderType: google_credentials` MCP servers).
-
-The OAuth token comes from a one-time interactive login. Run `agy` on the host
-and pick a login method:
-
-- **Google OAuth** -- personal Google accounts.
-- **Use a Google Cloud project** -- required for managed Workspace or corporate
-  domain accounts, which the plain OAuth path rejects with an auth error.
+agy authenticates from **gcloud ADC** (agy >= 1.1.10), which the harness opts
```
prernakakkar-google: These details should be left out for users.

evalbench/generators/models/agy_cli.py
```diff
@@ -198,53 +218,89 @@ def _ensure_agy_installed(self):
...
+            adc_path = GKE_SA_KEY_PATH
+            self.env["GOOGLE_APPLICATION_CREDENTIALS"] = adc_path
+
+        self.adc_path = adc_path if adc_path and os.path.exists(adc_path) else None
+        if self.adc_path:
+            logging.info("agy ADC resolved from %s", self.adc_path)
+            self._ensure_quota_project()
+        else:
+            # Not fatal: agy also resolves ADC from the GCE/GKE metadata
+            # server, which leaves nothing on disk to detect.
```
prernakakkar-google: This fallback doesn't hold up. A metadata server token has no quota project, and per your own docstring right below this, no quota project means an empty model registry and every --model gets rejected. So on a Workload Identity pod we land in exactly the state _ensure_quota_project exists to prevent, and the log tells whoever's debugging it that everything's fine and it fell back to the metadata server.

Before changing anything, have you tried GOOGLE_CLOUD_QUOTA_PROJECT? Not GOOGLE_CLOUD_PROJECT, which the docstring already rules out. The other one. We use it in scorers/dataformscorer.py:142 and the auth libraries take it over the file field. If agy respects it you'd fix this case and drop _ensure_quota_project and the credential copy entirely, which puts agy back in line with claude_code and codex_cli. Neither of those touches the credential at all.

If it doesn't work, make this a hard error saying a key file is required rather than a warning.

prernakakkar-google (dismissed):

evalbench/generators/models/agy_cli.py
omkargaikwad23: Tried GOOGLE_CLOUD_QUOTA_PROJECT. It doesn't work, agy models still comes back empty. Only the file field populates the registry. So _ensure_quota_project and the credential copy stay, and I've made the
  missing-file case a hard error as you suggested. Also did the same for the other two warn-and-continue branches so the method either guarantees the field or raises.

prernakakkar-google (approved):

PR #558, opened 2026-08-09
chore(viewer): delete the unused state.py

omkargaikwad23:
Nothing imports it. It is a stale copy of main.py's State: 16 fields
against 30, still carrying eval_summaries, and selected_main_tab defaults
to "List" rather than "Status".

Leaving it is a trap. @me.stateclass registers by class identity, so two
decorated State classes are two different state objects to Mesop. The
first module to import this one instead of main would get its own state
with its own defaults, silently, with no error to trace.

---

<sub>Stack created with <a href="https://github.com/github/gh-stack">GitHub Stacks CLI</a> • <a href="https://gh.io/stacks-feedback">Give Feedback 💬</a></sub>


omkargaikwad23:
Folded into #560 along with the other pure-cleanup changes, to keep the stack to one cleanup layer. Same diff, no content lost.

PR #563, opened 2026-08-10
Fix precompute trends oom

prernakakkar-google:
The Trends tab shows no data newer than **2026-06-25**. `trends_cache.csv` was last
written `2026-06-25T03:17:43Z`, holds 13,163 rows, and the newest `run_time` in it is
`2026-06-25 03:02:11`. 8,168 run directories have piled up behind it.

The precompute process is SIGKILLed roughly every 2.5 minutes. Every restart logs an
identical pair:

    WARN exited: precompute_trends (terminated by SIGKILL; not expected)
    Loaded 13163 processed directories from state.
    Found 21331 total directories. 8168 are new.

It then processes exactly 183 directories and dies. Because the old code wrote nothing
until the final lines of `precompute()`, all 183 were discarded every time — roughly
500 times a day for six weeks, with zero net progress.

### Root cause

`summarize_eval_scoring` pasted the entire `scores.csv` into the Gemini prompt via
`DataFrame.to_string()`. **That call pads every column to its widest value**, so a single
large `comparison_logs` cell is billed to every row in the frame.

Measured on `results/034b0627-59fa-4b4c-9193-612ed4a40304`:

| | |
|---|---|
| CSV on disk | 1.28 MB |
| Rows | **8** |
| Largest `comparison_logs` cell | 1,273,196 chars |
| `to_string()` output | **11,460,329 chars** = 1,536,138 tokens |
| Amplification | **8.9×**, almost entirely whitespace |

Gemini's limit is 1,048,576 tokens, so these were rejected outright — 61 rejections in a
45-minute window, ranging 1.5M to 4.36M tokens:

    400 INVALID_ARGUMENT: The input token count (1536138) exceeds the maximum
    number of tokens allowed (1048576).

Each rejection cost that run its summary and stored `"Error during summarization: ..."`
in the trends cache in its place. **50 of these strings were built concurrently** by
every precompute pass, which is what exhausted the container's memory.

Worth noting this is not a "too many rows" problem — the file above has 8 rows. Capping
row count would have changed nothing.

### DIFF
--- a/evalbench/test/viewer_precompute_trends_test.py
+++ b/evalbench/test/viewer_precompute_trends_test.py
@@ -1,0 +1,414 @@
+"""Tests for the viewer's trends precompute.
+
+The precompute runs against a results directory that grew past 20k run
+directories in production, where a pass that held everything in memory and only
+wrote at the end was killed before it ever finished. These tests pin the
+properties that keep that from recurring: work happens in bounded batches, each
+batch is durably checkpointed, and a kill part way through loses only the batch
+in flight.
+"""
+
+import json
+import os
+import sys
+import tempfile
+import unittest
+from unittest.mock import patch
+
+import pandas as pd
+
+# Add viewer directory and root to sys.path
+sys.path.append(os.path.abspath(os.path.join(os.path.dirname(__file__), "../../viewer")))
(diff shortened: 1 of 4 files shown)


IsmailMehdi (approved):

IsmailMehdi:
we should start switching the precompute to BQ instead of files.

PR #568, opened 2026-08-12
feat(dataset-quality): count skills in trajectory coverage

omkargaikwad23:
`trajectory_coverage` only looked at MCP tools. So a product that ships skills got
  graded against a surface its CUJs never name, and a skills-only product fell out of
  the grade entirely.

  The catalog is now every channel the product declares:

  - MCP tools and skill scripts, matched against `expected_trajectory`
  - skills themselves, matched against `expected_skills`

  New `generators/models/skills_catalog.py` reads `setup.skills` — clones a git URL or
  scans a local path, and takes the skill name from `SKILL.md` frontmatter. An entry
  that resolves to nothing raises instead of being skipped, since a short catalog
  quietly inflates coverage.

  Also renamed the sub-scorer metric keys — dropped the `dq_` prefix and fixed names
  that no longer matched what they count (`total_tools` → `capabilities_total`). They
  live inside a JSON blob, so nothing downstream changes.

  ## Test plan
  - [x] `pytest test/dataset_quality_test.py -q` — 88 passed
  - [x] Live Cloud SQL run: **63 (C)**, graded on 100/100 weight, no excluded scorers
  - [x] Tested all internal MCP servers and skills agaist available CUJs
  - [x] Config declaring both channels: coverage went `0/15` → `5/63`

### DIFF
--- a/datasets/dataset_quality/example_run_config.yaml
+++ b/datasets/dataset_quality/example_run_config.yaml
@@ -1,9 +1,8 @@
 ############################################################
 ### Dataset Quality — example run config
 ############################################################
-# Grades a product's CUJ configs, not a model: no agent runs and nothing is
-# executed. The product itself is only read, to supply the capability surface
-# those CUJs are scored against.
+# Grades the dataset's CUJs, not a model: no agent runs and nothing is executed.
+# The product only supplies the tools and skills those CUJs are scored against.
 
 # Forced by dataset_format below; listed here only to document what runs.
 orchestrator: agent
--- a/datasets/dataset_quality/example_run_config.yaml
+++ b/datasets/dataset_quality/example_run_config.yaml
@@ -1,10 +1,9 @@
 ############################################################
 ### Dataset Quality — example run config
 ############################################################
-# Grades the intrinsic quality of a CUJ dataset (no eval / agent run). Runs on the
-# standard agent pipeline: a no-op generator does nothing, and the `dataset_quality`
-# scorer grades the whole dataset holistically (each sub-scorer computes a 0-100
-# sub-score; grading rolls them into a weighted global score + letter grade).
+# The dataset itself is what's under test here, not a model: no agent runs and
+# nothing is executed. The product is only read, for the capabilities to grade
+# the dataset's coverage against.
 
 # Forced by dataset_format below; listed here only to document what runs.
 orchestrator: agent
@@ -26,12 +25,9 @@ simulated_user_model_config: datasets/model_configs/noop_agent_model.yaml
 ############################################################
 ### Scorers
 ############################################################
-# The `dataset_quality` scorer fetches the product's tool catalog live from its own
-# model_config (each setup.mcp_servers entry with an httpUrl/url is queried via MCP
-# tools/list; names are namespaced <server>__<tool> to match expected_trajectory).
-# google_credentials servers use ADC, so this needs credentials + network (e.g.
-# `gcloud auth application-default login`). Run the LLM judges with
-# EVAL_GCP_PROJECT_REGION=global.
+# Reads the product's capability surface from the model_config below:
+# setup.mcp_servers for tools (named <server>__<tool>, matching
+# expected_trajectory) and setup.skills for skills.
 scorers:
   dataset_quality:
     model_config: datasets/model_configs/gemini_cli_model.yaml
@@ -41,8 +37,10 @@ scorers:
     # required only by the judge sub-scorers; trajectory_coverage and
     # naming_distribution are static.
     sub_scorers:
+      # Scores the finest catalog the product exposes: MCP tools, else the
+      # scripts its skills ship, else the skills themselves.
       trajectory_coverage:
-        # weight: 20  (default)
+        # weight: 30  (default)
       naming_distribution:
         # weight: 5   (default)
       parameter_coverage:
@@ -71,8 +69,6 @@ scorers:
 ############################################################
 ### Reporting
 ############################################################
-# The grade lands as one summary scoring row plus one row per category; uncomment
-# bigquery to also append to the table.
 reporting:
   csv:
     output_directory: 'results'
--- a/datasets/dataset_quality/example_run_config.yaml
+++ b/datasets/dataset_quality/example_run_config.yaml
@@ -37,10 +37,10 @@ scorers:
     # required only by the judge sub-scorers; trajectory_coverage and
     # naming_distribution are static.
     sub_scorers:
-      # Scores the finest catalog the product exposes: MCP tools, else the
-      # scripts its skills ship, else the skills themselves.
+      # Scores every channel the product exposes at once: MCP tools and the
+      # scripts its skills ship, plus the skills themselves.
       trajectory_coverage:
-        # weight: 30  (default)
+        # weight: 20  (default)
       naming_distribution:
         # weight: 5   (default)
       parameter_coverage:
--- a/datasets/dataset_quality/example_run_config.yaml
+++ b/datasets/dataset_quality/example_run_config.yaml
@@ -1,9 +1,9 @@
 ############################################################
 ### Dataset Quality — example run config
 ############################################################
-# The dataset itself is what's under test here, not a model: no agent runs and
-# nothing is executed. The product is only read, for the capabilities to grade
-# the dataset's coverage against.
+# Grades a product's CUJ configs, not a model: no agent runs and nothing is
+# executed. The product itself is only read, to supply the capability surface
+# those CUJs are scored against.
 
 # Forced by dataset_format below; listed here only to document what runs.
 orchestrator: agent
@@ -37,8 +37,8 @@ scorers:
     # required only by the judge sub-scorers; trajectory_coverage and
     # naming_distribution are static.
     sub_scorers:
-      # Scores every channel the product exposes at once: MCP tools and the
-      # scripts its skills ship, plus the skills themselves.
+      # Counts MCP tools and skill scripts (matched against a CUJ's
+      # expected_trajectory) plus each skill's own name (expected_skills).
       trajectory_coverage:
         # weight: 20  (default)
       naming_distribution:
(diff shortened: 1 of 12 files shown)


datasets/dataset_quality/example_run_config.yaml
```diff
@@ -1,10 +1,9 @@
 ############################################################
 ### Dataset Quality — example run config
 ############################################################
-# Grades the intrinsic quality of a CUJ dataset (no eval / agent run). Runs on the
-# standard agent pipeline: a no-op generator does nothing, and the `dataset_quality`
-# scorer grades the whole dataset holistically (each sub-scorer computes a 0-100
-# sub-score; grading rolls them into a weighted global score + letter grade).
+# The dataset itself is what's under test here, not a model: no agent runs and
+# nothing is executed. The product is only read, for the capabilities to grade
```
prernakakkar-google: CUJ configs of each product

datasets/dataset_quality/example_run_config.yaml
```diff
@@ -41,6 +37,8 @@ scorers:
     # required only by the judge sub-scorers; trajectory_coverage and
     # naming_distribution are static.
     sub_scorers:
+      # Scores every channel the product exposes at once: MCP tools and the
+      # scripts its skills ship, plus the skills themselves.
```
prernakakkar-google: Comment not very clear

evalbench/generators/models/skills_catalog.py
```diff
@@ -0,0 +1,258 @@
...
+class Skill:
+    """One skill in a product's catalog.
+
+    A skill groups operations rather than being one: `scripts` holds the files in
+    its adjacent scripts/ directory, which is what a CUJ trajectory names.
+    """
+
+    name: str
+    description: str = ""
+    scripts: tuple[str, ...] = ()
```
prernakakkar-google, line 45: SCripts is for MCP toolbox generated skills but its not necessary that will be the case.

evalbench/generators/models/skills_catalog.py
```diff
@@ -0,0 +1,258 @@
...
+        return _scan_dir(clone_dir, wanted)
+    finally:
+        shutil.rmtree(clone_dir, ignore_errors=True)
+
+
+def _clone(url: str) -> str:
+    """Clones a skills repository into a temporary directory (supports '<url>#<ref>')."""
+    clone_url, _, ref = url.partition("#")
+    dest = tempfile.mkdtemp(prefix="skills_catalog_")
+    base = ["git", "clone", "--depth", "1"]
```
prernakakkar-google: git clone --depth 1 --branch won't accept a commit SHA, so a SHA-pinned #ref fails the first attempt and this falls through to a default-branch clone — we grade against the wrong catalog with no signal. codex_cli._clone_extension_repo deliberately doesn't retry. Given this module's fail-loud stance, I'd drop the fallback and let the ref failure raise. The current test mocks a clone that never fails, so this path is uncovered.

evalbench/generators/models/skills_catalog.py
```diff
@@ -0,0 +1,258 @@
...
+            )
+            return dest
+        except (subprocess.CalledProcessError, subprocess.TimeoutExpired,
+                OSError) as e:
+            last_error = e
+    shutil.rmtree(dest, ignore_errors=True)
+    raise SkillCatalogError(f"clone failed for {url}: {last_error}")
+
+
+def _scan_dir(root: str, wanted: set[str] | None = None) -> list[Skill]:
```
prernakakkar-google, line 133: This raises only when zero wanted names match. With skills: ["alpha", "typo", "gamma"] against a root holding alpha/beta/gamma, it returns 2 and stays silent — which is exactly the shrunken denominator the module docstring says it prevents. Can we raise on any unmatched name in wanted, and extend test_a_narrowing_key_matching_nothing_is_fatal to the partial-miss case?

evalbench/scorers/dataset_quality/scorer.py
```diff
@@ -125,43 +126,29 @@ def compare(
...
-                    {
-                        "product_name": self.product_name,
-                        "graded": False,
-                        "error": f"tool discovery failed: {e}",
-                    },
-                    default=str,
-                ),
-            )]
+            tools = self._fetch_tools(setup)
+            skills = resolve_skills(setup)
```
prernakakkar-google: A skills-resolution failure now nulls the entire report, including the five judge sub-scorers that never touch skills. gemini_cli_skills_model.yaml:12 hardcodes an absolute path under one person's home dir — on any other machine that config goes from "grades the tools" to "ungraded." Should a skills failure degrade skill coverage rather than the whole grade?

prernakakkar-google:
The denominator quadruples (0/15 → 5/63 in your own test), so dataset_quality scores for skills-bearing products aren't comparable across this change. What's the plan for the step change on the trends page — new series, or annotate the boundary? Same question for the dq_* rename: code is fine (the viewer renders metrics generically), but historical BigQuery rows keep the old keys, so any saved query over the blob breaks at the boundary.

prernakakkar-google (changes requested):

evalbench/generators/models/skills_catalog.py
omkargaikwad23: Makes sense. Reworded to say scripts/ is the MCP-toolbox convention and that a skill without one is graded on its name alone.

evalbench/generators/models/skills_catalog.py
omkargaikwad23: Good catch. Any unmatched name now raises with all the missing ones listed, and added a test for the partial-miss case.

evalbench/generators/models/skills_catalog.py
omkargaikwad23: Good catch on the SHA. --branch only takes a branch or tag, so a SHA-pinned ref was quietly landing on the default branch. Dropped the fallback: one attempt, and a ref that doesn't resolve now raises.

evalbench/scorers/dataset_quality/scorer.py
omkargaikwad23: There's no separate skills sub-scorer to degrade, trajectory_coverage grades tools, scripts and skills as one union catalog, so a skills failure would shrink denominator and inflate the score rather lower it. 
On gemini_cli_skills_model.yaml: that absolute path is worth fixing but it's separate file config change.

omkargaikwad23:
> The denominator quadruples (0/15 → 5/63 in your own test), so dataset_quality scores for skills-bearing products aren't comparable across this change. What's the plan for the step change on the trends page — new series, or annotate the boundary? Same question for the dq_* rename: code is fine (the viewer renders metrics generically), but historical BigQuery rows keep the old keys, so any saved query over the blob breaks at the boundary.

Thanks for the review @prernakakkar-google! 
Only  2 products declare skills today, and one each, so the denominator barely moves for them. Tools-only  products are unaffected -> same catalog, same score. 

For the dq_ rename, nothing consumes those keys by name; the viewer renders whatever the blob holds.

datasets/dataset_quality/example_run_config.yaml
omkargaikwad23: Updated.

datasets/dataset_quality/example_run_config.yaml
omkargaikwad23: Updated!

datasets/dataset_quality/example_run_config.yaml
```diff
@@ -1,10 +1,9 @@
 ############################################################
 ### Dataset Quality — example run config
 ############################################################
-# Grades the intrinsic quality of a CUJ dataset (no eval / agent run). Runs on the
-# standard agent pipeline: a no-op generator does nothing, and the `dataset_quality`
-# scorer grades the whole dataset holistically (each sub-scorer computes a 0-100
-# sub-score; grading rolls them into a weighted global score + letter grade).
+# Grades a product's CUJ configs, not a model: no agent runs and nothing is
```
prernakakkar-google: Can we simplify this to say It helps with Datset CUJ grading.

evalbench/generators/models/skills_catalog.py
```diff
@@ -0,0 +1,264 @@
...
+def _is_script(filename: str) -> bool:
+    """Whether a scripts/ entry is something a CUJ could name.
+
+    A file no trajectory can ever name still lands in the coverage denominator,
+    so it caps the achievable score and is reported as a permanent gap.
+    """
+    stem, suffix = os.path.splitext(filename)
+    if filename.startswith(".") or filename == "__init__.py":
+        return False
+    if suffix.lower() in _NON_SCRIPT_SUFFIXES:
```
prernakakkar-google: I think it would be more accurate  to check in given list of .js, .ts, .py, .sh

evalbench/scorers/dataset_quality/scorer.py
prernakakkar-google: Agreed that silently shrinking the catalog is worse than failing. But the degradable unit is trajectory_coverage, not a separate skills scorer: _run_scorers already catches a sub-scorer exception, marks it applicable=False, and grading.py reports it in excluded_scorers with the remaining weight. Moving resolve_skills into TrajectoryCoverageScorer.run would drop just that scorer's 20 weight on a bad catalog and still grade the five judges — no inflated denominator, no lost report. As it stands a bad setup.skills path nulls vague_examples, cuj_diversity, error_recovery, composition, and parameter_coverage, none of which touch the catalog.

prernakakkar-google (dismissed):

prernakakkar-google:
Few minor comments, otherwise LGTM

datasets/dataset_quality/example_run_config.yaml
omkargaikwad23: Makes sense!

evalbench/scorers/dataset_quality/scorer.py
omkargaikwad23: Good call. Split the try. Tool discovery still nulls the report since every sub-scorer reads tool_names, but a SkillCatalogError now just marks trajectory_coverage inapplicable, so it lands in excluded_scorers and the other 80 weight grades as normal. Added a test asserting exactly that.

datasets/dataset_quality/example_run_config.yaml
omkargaikwad23: Done.

helloeve (approved):

PR #569, opened 2026-08-12
Allow per-dialect schema files

aseering:
Spanner has two distinct dialects (PostgreSQL and GoogleSQL):  The engine is functionally the same and the dataset is the same, but the schema .sql files need to be different per-database.  This change adds automatic (but optional) support for having per-dialect schema files in a dataset.


prernakakkar-google (approved):

aseering:
yep, still not a collaborator.  Would someone who is, mind kicking off the tests here?

prernakakkar-google:
> yep, still not a collaborator. Would someone who is, mind kicking off the tests here?

Added you as collaborator. Thanks Adam for contributions.

PR #577, opened 2026-08-19
feat: Add support for eval case level timeouts

twang126:
Adds eval-case level timeout support across orchestrators, evaluators, and agent CLI generators.


prernakakkar-google:
Please fix tests and pycodestyle

evalbench/evaluator/agentevaluator.py
```diff
@@ -4,12 +4,14 @@
 import logging
 import os
 import shutil
+import time
```
prernakakkar-google, line 7: Can you add example run config and attach gpaste of run logs

evalbench/evaluator/agentevaluator.py
```diff
@@ -238,14 +286,20 @@ def _finalize_scenario(
         metadata: Dict[str, Any]
     ):
         """Finalizes the scenario by scoring and appending results."""
+        generated_error = None
+        if hasattr(last_result, "returncode") and last_result.returncode != 0:
+            generated_error = getattr(last_result, "stderr", None) or f"Command failed with exit code {last_result.returncode}"
```
prernakakkar-google: This sets generated_error on any non-zero exit, not just timeouts, and agentscorework.py:54 now passes it into score.compare. Agent CLIs often exit non-zero but still return good output — agy_cli.py:919 says exactly that in a comment. I ran a scenario with no eval_case_timeout set anywhere: main gives generated_error=None, this branch gives the stderr string, and tool_call_latency, token_consumption, effective_billed_tokens and agent_steps all drop to 0.0. That moves scores on existing evalsets

evalbench/evaluator/evaluator.py
```diff
@@ -71,8 +72,12 @@ def __init__(
         self.sqlgen_runners = runner_config.get("sqlgen_runners", 10)
         self.sqlexec_runners = runner_config.get("sqlexec_runners", 10)
         self.scoring_runners = runner_config.get("scoring_runners", 10)
-        self.task_timeout_seconds = runner_config.get(
-            "task_timeout_seconds", 600)
+        eval_timeout = get_eval_case_timeout(self.config)
+        self.task_timeout_seconds = (
+            eval_timeout
+            if eval_timeout is not None
+            else runner_config.get("task_timeout_seconds", 600)
```
prernakakkar-google: task_timeout_seconds isn't a per-case budget. It's a stall watchdog — the clock resets whenever any future finishes (lines 32-34) — and it's used for promptgen, sqlgen, sqlexec and scoring (121, 160, 213, 235, 283). So eval_case_timeout: 30s would drop all pending futures after 30s with no completion, which scoring alone can hit on a large dataset. AgentEvaluator never reads task_timeout_seconds at all, so this only lands on the SQL path. Suggest: keep the two keys separate

evalbench/evaluator/agentevaluator.py
twang126: put the config, some execution logs, and the resulting csv here: https://paste.googleplex.com/5020206278770688

twang126: I have also added the example config to this PR and udpated the documentation with the new setting.

prernakakkar-google: I see the resultant scorers are 0 as per gpaste.

prernakakkar-google:
For gemini on timeout returncode is 124, so the raw NDJSON stream is never converted. Downstream parse_response hits JSONDecodeError, returns {}, and extract_tools/extract_skills return [] — the entire partial trajectory is lost.

evalbench/evaluator/agentevaluator.py
```diff
@@ -238,12 +286,18 @@ def _finalize_scenario(
         metadata: Dict[str, Any]
     ):
         """Finalizes the scenario by scoring and appending results."""
+        timed_out = (
```
prernakakkar-google, line 289: Any agent that surfaces a Python traceback from an MCP tool, or a CLI that logs its own internal API TimeoutError, gets flagged as a harness timeout despite completing normally.

evalbench/evaluator/agentevaluator.py
```diff
@@ -205,18 +249,22 @@ def process_scenario(
...
-                scenario,
-                last_result,
-                conversation_history,
-                accumulated_tools,
-                accumulated_skills,
-                eval_result,
-                job_id,
-                metadata
+        if last_result is None:
+            last_result = subprocess.CompletedProcess(
```
prernakakkar-google, line 253: This is a behaviour change outside the timeout feature: a scenario with max_turns: 0 was previously skipped and is now scored as a successful empty run.

evalbench/evaluator/agentevaluator.py
twang126: Fixed, and added unit test

evalbench/evaluator/agentevaluator.py
twang126: Reverted

prernakakkar-google (dismissed):

prernakakkar-google (approved):

PR #578, opened 2026-08-19
feat(analytics-scorer): add AnalyticsScorer implementation and update…

MohammadrezaPourreza:
Add the AnalyticsScorer as a new scorer.


evalbench/scorers/analyticsscorer.py
```diff
@@ -0,0 +1,146 @@
...
+        generated_query: Any,
+        generated_execution_result: Any,
+        generated_eval_result: Any,
+        generated_error: Any,
+        database: str = "",
+        **kwargs,
+    ) -> Tuple[float, str]:
+        """Evaluates trial result against golden reference following the Brewmax rubric."""
+        has_golden_data = bool(golden_execution_result)
+        has_generated_data = bool(generated_execution_result)
```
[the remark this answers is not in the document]
MohammadrezaPourreza: Fixed

evalbench/test/analyticsscorer_test.py
```diff
@@ -0,0 +1,136 @@
...
+    def test_render_data_truncation(self, mock_get_gen):
+        mock_get_gen.return_value = MagicMock()
+        scorer = AnalyticsScorer(
+            {"model_config": "model.yaml", "max_data_chars": 50},
+            global_models={},
+        )
+        long_data = [{"id": i, "description": "some long text field value"} for i in range(10)]
+        rendered = scorer._render_data(long_data)
+        self.assertIn("... [truncated ", rendered)
+        self.assertTrue(len(rendered) < 150)
```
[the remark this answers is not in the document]
MohammadrezaPourreza: Fixed.

evalbench/scorers/analyticsscorer.py
```diff
@@ -0,0 +1,145 @@
...
+        database: str = "",
+        **kwargs,
+    ) -> Tuple[float, str]:
+        """Evaluates trial result against golden reference following the Brewmax rubric."""
+        has_golden_data = bool(golden_execution_result)
+
+        # Handle execution error conditions per rubric
+        if golden_error:
+            return 0.0, f"Golden query failed to execute: {golden_error}"
+        if generated_error and has_golden_data:
```
IsmailMehdi: **Blocking: a generated query that errored scores 100 when the golden result is empty.**

This guard misses the case where the golden result is legitimately empty. With `has_golden_data == False` the errored generation falls through to the LLM, which sees no data on either side and passes it. Verified against this branch:

```
compare(..., golden_execution_result=[], generated_error="Syntax error: no such table")
-> (100.0, 'VERDICT: PASS')   # LLM called: True
```

A query that failed to execute is not equivalent to one that legitimately returned zero rows. `llmrater.py` returns `0` on any `generated_error`. Suggest dropping the `and has_golden_data` so the guard is unconditional -- otherwise this inflates scores on any dataset with empty-result golden queries.

evalbench/scorers/analyticsscorer.py
```diff
@@ -0,0 +1,145 @@
...
+            data_str = ""
+        elif isinstance(data, (list, dict)):
+            try:
+                data_str = json.dumps(data, default=str)
+            except Exception:
+                data_str = str(data)
+        else:
+            data_str = str(data)
+
+        if len(data_str) > self.max_data_chars:
```
IsmailMehdi: **Blocking: character-level truncation hands the judge corrupted records.**

Cutting the serialized JSON at a character offset splits mid-record. Verified with `max_data_chars=60`:

```
'[{"name": "Alice", "salary": 100000}, {"name": "Bob", "salar... [truncated 11 chars]'
```

The judge receives a half-written record and unbalanced brackets. Rubric point 9 explicitly asks the model to reason about truncation, so malformed input directly corrupts the thing being graded -- and at the 8000-char default this fires on any realistically-sized result set.

`llmrater` solves this with `take_n_uniques(result, 50)`, which samples at row level and always yields well-formed records. Suggest truncating by rows and stating the omitted row count.

evalbench/scorers/analyticsscorer.py
```diff
@@ -0,0 +1,145 @@
...
+
+        cleaned = response_text.strip()
+        matches = re.findall(r"VERDICT:\s*(PASS|FAIL)\b", cleaned, re.IGNORECASE)
+        if matches:
+            verdict = matches[-1].upper()
+            score = 100.0 if verdict == "PASS" else 0.0
+            return score, response_text
+
+        # Fallback heuristic if explicit label prefix is omitted
+        if "PASS" in cleaned.upper() and "FAIL" not in cleaned.upper():
```
IsmailMehdi: **Blocking: substring matching on `PASS` produces false passes.**

Verified against this branch:

| Response | Score |
|---|---|
| `"The trial result passes the rubric."` | **100.0** |
| `"This surpasses the requirement."` | **100.0** |
| `"Rating: 1.0. Correct."` | 0.0 |

Both `passes` and `surpasses` trigger a pass. This fallback only runs when the model ignored the output format -- i.e. exactly when its output is least trustworthy -- so defaulting to `0.0` with a "could not parse verdict" log would be safer, and would surface prompt-compliance problems instead of hiding them behind a coin flip.

(The primary `re.findall(...)[-1]` path above is well chosen; taking the last match is the right call.)

evalbench/scorers/prompt/analyticsscorer.py
```diff
@@ -0,0 +1,39 @@
...
+
+The rubric is not satisfied if according to the above criteria, the trial data result does not correctly address the user prompt."""
+
+ANALYTICS_SCORER_PROMPT_TEMPLATE = """Your task is to check how Conversational Analytics agent trial responses to a user prompt compare to ground truth Conversational Analytics responses for a single conversational turn.
+Note that the ground truth responses serve as a reference for the trial responses, not a strict template that must be matched exactly.
+Below is the rubric which determines how to evaluate trial responses given ground truth responses.
+
+For each turn, based on the above rubric, the rater will return a rating given three types of information: 1) a user prompt that responses attempt to address, 2) the ground truth responses to address the user prompt, and 3) the trial responses to address the user prompt.
+When returning a rating that is less than 1.0, the rater must justify this rating by clearly documenting the shortcomings of the trial response in the provided explanation.
```
IsmailMehdi: **Blocking: the template gives the model three conflicting output contracts.**

- This line and "the rater will return a rating" above imply a continuous 0-1 score.
- Line 25 ("Output ratings corresponding to the rubric for each property, like as follows:") promises a format example, then goes straight to `# Rubric` -- the example is missing.
- Line 39 states the actual contract: `VERDICT: PASS` or `VERDICT: FAIL`, which is the only thing parsed.

Also, line 20's "Below is the rubric" precedes a paragraph, not the rubric -- `{rubric}` doesn't appear until line 27.

Given that the fallback heuristic at `analyticsscorer.py:75` exists precisely to catch format non-compliance, tightening this prompt is the higher-leverage fix. Suggest dropping the 1.0-rating language and the dangling "as follows:", and keeping only the PASS/FAIL contract.

IsmailMehdi (commented):
Reviewed by running the branch locally and probing the scoring edge cases. The framework integration is clean -- `__init__(config, global_models)` matches what `score._build_instances` introspects for, the caching path mirrors `llmrater`, and the 0/100 scale is consistent with the other comparators. All 14 tests pass.

Leaving inline notes on four blocking items, all in the scoring semantics. The first one silently awards 100 to queries that failed to execute.

Non-blocking items I'd also raise, but didn't inline:

- **Swallowing exceptions converts infra failures into real zeros.** `analyticsscorer.py:143` catches `Exception` and returns `(0.0, ...)`, but `score.compare()` already wraps every comparator and records the exception in `ComparisonResult.comparison_error`. Catching here writes an auth/quota/timeout failure to the report as a legitimate score of 0 with `comparison_error=None`, indistinguishable from a genuine FAIL. A partial outage would silently depress a whole run. `llmrater` deliberately doesn't catch.
- **No cheap short-circuit.** `llmrater` runs `SetMatcher` first and skips the model on an exact match. This scorer calls the judge on every row, which is a real cost/latency increase for the one case where the LLM adds nothing.
- **`getattr(response, "stdout", response)` (line 83) is dead code.** Nothing in the generator hierarchy returns an object with `.stdout` from `generate()`; the judge generators return `str`.
- **`viewer/version.txt` is an unrelated build artifact.** It moves to `797b8c6` (the merge commit of #574, a dependabot bump) picked up by a local build. `main` is already past that point, so merging rolls the file backwards. Worth reverting from this PR.
- **"Brewmax" appears in four docstrings** (module, class, prompt module, test). Worth confirming that internal codename is cleared for a public `GoogleCloudPlatform` repo, otherwise "Conversational Analytics Data Results Rater" as used in `docs/scorers.md` reads better.
- Minor: `docs/scorers.md` new row is a run-on missing a period ("...Data Results Rater Uses an LLM judge..."); `_render_trajectory`'s `query_label` default is never used since both call sites pass it explicitly.

On tests: good structural coverage, and the `assertIn` checks on the generated prompt are a nice touch. The gaps map onto the bugs below, which is why they survived -- no case for `generated_error` with an empty golden result, none for a response lacking the `VERDICT:` label (the only path that runs the fallback), none for the model raising, and `test_render_data_truncation` asserts the marker and a length bound but never that the surviving payload is parseable.

No security concerns: no new credential handling, no shell or SQL construction, no untrusted deserialization.

IsmailMehdi (dismissed):

evalbench/scorers/analyticsscorer.py
MohammadrezaPourreza: fixed.

evalbench/scorers/analyticsscorer.py
MohammadrezaPourreza: fixed.

evalbench/scorers/analyticsscorer.py
MohammadrezaPourreza: fixed.

evalbench/scorers/prompt/analyticsscorer.py
MohammadrezaPourreza: fixed.

IsmailMehdi (approved):

PR #579, opened 2026-08-24
Fix evalbench dashboard outage: render timeouts, precompute crash-loop, duplicate precompute writers

prernakakkar-google:
**Summary**
The dashboard at evalbench-463971716997.cr.gclb.goog was returning 500 on every /__ui__ render (120s, the request timeout). Three independent causes, all fixed here. Post-deploy it renders in ~1.8–3.3s.

Root causes
1. The render path stat'd the whole results mount, three times per page.
Each render ran [d for d in os.listdir(p) if os.path.isdir(...)] in three places. os.path.isdir can't reuse anything os.listdir learned, so it costs one FUSE round trip per entry — ~25k entries × 3 walks ≈ 75k round trips per render, which alone exceeded the timeout.

2. Precompute was crash-looping and had processed 695 of 24,969 runs.
A run whose summary.csv lacked metric_name raised past the batch loop without being added to batch_processed, so it was rescanned on every pass, forever. The backlog never drained, and the cache the UI depends on never grew.

3. Every Cloud Run instance ran its own precompute against the same bucket.
maxScale=20, and supervisord starts the precompute loop in every instance — five were observed running identical passes over the same 22k directories. Beyond ~5× wasted work (dominated by one Gemini call per run), both caches are rewritten whole, so a later write silently discards an earlier one's rows while processed_dirs.json still records those runs as done. Anything lost that way is never reconsidered and never appears in trends again.

**Changes**
1. viewer/run_index.py (new) — shared os.scandir-based listing with a 60s TTL cache. scandir carries the dirent type through, so is_dir() needs no second trip. Callers in main.py, trends.py, and precompute_dataset_quality.py; the clear-cache handler invalidates it.
2. viewer/precompute_trends.py — return None was overloaded, meaning both "artifacts not written yet, retry" and "this threw"; the second is what jammed the queue. Now split: ParserError/EmptyDataError/OSError still defer (a run caught mid-write or a transient FUSE read deserves a retry), while anything else raises UnparsableRun and is quarantined as processed. Also fixes a second instance of the same bug on a missing run_time column.
3. viewer/precompute_lease.py (new) + run_precompute.py — single-writer lease. Acquiring writes a holder ID and wall-clock timestamp, waits out the competing-write window, then reads back; a GCS whole-file rewrite has exactly one winner and reads are strongly consistent after it, so whoever reads their own ID back holds it. Renewed every 30s against a 120s TTL, so long passes keep the lease but a killed holder's lease is taken over — which matters, since a large backlog can get precompute SIGKILLed mid-pass. Fails open: if the lease can't be written, precompute runs unguarded, because a dashboard that silently stops updating is worse than duplicated work.
4. ai_summary windowing — it was 95.8% of trends_cache.csv, on track for ~93MB read on every render once the backlog drained. Now kept inline only for the last 7 days and written to a per-run ai_summary.txt sidecar, which the detail view reads. Existing summaries are backfilled to sidecars before being blanked (~3.3MB → 209KB measured; ~93MB → ~5MB projected).
5. supervisord_cloudrun.conf — PRECOMPUTE_WORKERS=12, PRECOMPUTE_INTERVAL=600. Precompute stays in the serving container so trends keep auto-updating.


viewer/precompute_lease.py
```diff
@@ -0,0 +1,140 @@
...
+        self.acquired = False
+        self._stop = threading.Event()
+        self._renewer = None
+
+    def acquire(self):
+        holder, timestamp = _read(self.results_dir)
+        age = time.time() - timestamp
+
+        if holder and holder != _HOLDER_ID and age < LEASE_TTL_SECONDS:
+            logging.info(
```
helloeve, line 69: non-blocking comment: where can I inspect these logs?

viewer/precompute_trends.py
```diff
@@ -14,6 +19,45 @@
 # where the summarizer starts seeing 429s, so raising it further buys nothing.
 MAX_WORKERS = int(os.environ.get("PRECOMPUTE_WORKERS", 50))
 
+# ai_summary is ~96% of trends_cache.csv, which is read in full on every render.
+# Only the recent window stays inline; every summary is also written beside its
+# own run so nothing is lost.
+AI_SUMMARY_FILENAME = "ai_summary.txt"
+AI_SUMMARY_CACHE_DAYS = int(os.environ.get("AI_SUMMARY_CACHE_DAYS", 7))
```
helloeve, line 26: just trying to think out loud here -- how about having this display-as needed so it is not preloaded?

helloeve (approved):
overall LGTM! Left a few non-blocking comments.

viewer/precompute_lease.py
prernakakkar-google: You can find this in Cloud run logs:
https://pantheon.corp.google.com/run/detail/us-central1/evalbench/observability/logs?e=13802955&mods=-autopush_coliseum&project=evalbench-dev

viewer/precompute_trends.py
prernakakkar-google: I think that sounds better approach let me do this in follow up PR.

PR #583, opened 2026-08-28
fix(mcp): use configured MCP server URL and unwrap taskgroup errors

omkargaikwad23:
## Description
- **Preserve configured URLs**: Removed URL alteration logic in `mcp_client.py` that forced an `/mcp` suffix onto every endpoint. Configured URLs are now trusted directly (with whitespace trimmed), allowing endpoints with custom paths (e.g. `https://<host>/mcp/<service>` or `https://<host>/api/v1/mcp`) to work without being broken by redundant path appending.
- **Unwrap TaskGroup errors**: Added a compact `_format_error` helper in `mcp_client.py` to extract leaf errors from nested `ExceptionGroup` / `TaskGroup` exceptions, replacing the generic `"unhandled errors in a TaskGroup (1 sub-exception)"` with actionable error causes in reports and UI views.

## Testing
- Verified live capability discovery against Streamable HTTP MCP endpoints with custom subpaths.
- Ran pytest across test suite (122 passed).
- Verified style check with pycodestyle (0 errors).
- Verified all internal products: https://screenshot-v2.corp.google.com/1fg3a2cfjaq2g

### DIFF
--- a/evalbench/generators/models/mcp_client.py
+++ b/evalbench/generators/models/mcp_client.py
@@ -20,23 +20,11 @@ class McpToolsError(Exception):
     """Raised when a tools spec cannot be fetched or parsed."""
 
 
-def sanitize_url(url: str) -> str:
-    """Ensure the URL has a scheme and ends with ``/mcp``.
-
-    ``/mcp`` is the conventional path where the MCP protocol is exposed.
-    """
-    stripped_url = (url or "").strip()
-    if stripped_url.startswith(("http://", "https://")):
-        url_with_scheme = stripped_url
-    elif "localhost" in stripped_url:
-        url_with_scheme = f"http://{stripped_url}"
-    else:
-        url_with_scheme = f"https://{stripped_url}"
-
-    rstripped_url = url_with_scheme.rstrip("/")
-    if not rstripped_url.endswith("/mcp"):
-        return f"{rstripped_url}/mcp"
-    return rstripped_url
+def _format_error(e: BaseException) -> str:
+    """Unpack ExceptionGroup/TaskGroup leaf errors for error messages."""
+    if hasattr(e, "exceptions") and getattr(e, "exceptions"):
+        return f"{e} ({', '.join(map(_format_error, e.exceptions))})"
+    return f"{type(e).__name__}: {e}" if str(e) else type(e).__name__
 
 
 def auth_headers(server_config: dict) -> dict | None:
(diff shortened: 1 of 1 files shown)


anubhav756 (approved):

anubhav756 (approved):

evalbench/generators/models/mcp_client.py
```diff
@@ -20,23 +20,11 @@ class McpToolsError(Exception):
     """Raised when a tools spec cannot be fetched or parsed."""
 
 
-def sanitize_url(url: str) -> str:
```
wangauone, line 23: We have the same function in here https://github.com/GoogleCloudPlatform/evalbench/blob/main/evalbench/generators/models/mcp_tools.py#L93-L109. Do we also want to make some similar changes there as well?

evalbench/generators/models/mcp_client.py
```diff
@@ -20,23 +20,11 @@ class McpToolsError(Exception):
...
-    elif "localhost" in stripped_url:
-        url_with_scheme = f"http://{stripped_url}"
-    else:
-        url_with_scheme = f"https://{stripped_url}"
-
-    rstripped_url = url_with_scheme.rstrip("/")
-    if not rstripped_url.endswith("/mcp"):
-        return f"{rstripped_url}/mcp"
-    return rstripped_url
+def _format_error(e: BaseException) -> str:
```
wangauone, line 23: nit: Ideally, it would be good to have a mcp_client_test.py to do some testing. : ) But I think it is okay to limit the scope of this PR.

evalbench/generators/models/mcp_client.py
```diff
@@ -20,23 +20,11 @@ class McpToolsError(Exception):
...
 
-def sanitize_url(url: str) -> str:
-    """Ensure the URL has a scheme and ends with ``/mcp``.
-
-    ``/mcp`` is the conventional path where the MCP protocol is exposed.
-    """
-    stripped_url = (url or "").strip()
-    if stripped_url.startswith(("http://", "https://")):
-        url_with_scheme = stripped_url
-    elif "localhost" in stripped_url:
```
wangauone, line 31: We will lost the capability to parse the value like `localhost:8080` or `sqladmin.googleapis.com` (value without http prefix)? Just want to make sure this is the expected behavior

wangauone (approved):
LGTM with some minor comments

evalbench/generators/models/mcp_client.py
omkargaikwad23: Sg! I will create a followup PR to add required test files.

evalbench/generators/models/mcp_client.py
omkargaikwad23: That's correct. We are assuming the fully qualified URL will be provided for now. We can definitely revisit this and add support for scheme-less URLs in the future if needed

evalbench/generators/models/mcp_client.py
omkargaikwad23: That's out of scope for now as those files are related to mcp-readability flow. Maybe I can check with owner of the file.

PR #584, opened 2026-08-28
feat(agy_cli): pass --add-dir flag automatically on agy CLI invocations

silviojr:
When executing Antigravity (`agy`) CLI scenarios, `agy` expects active workspace directories to be registered. It seems that when it executes in an empty folder (fake_home), it just ignores the directory as a workspace directory. 

However, there's the `--add-dir` command-line flag. With this flag, `agy` initializes in non-interactive print mode with a registered workspce.

This PR updates `AgyCliGenerator` to automatically pass `--add-dir <cwd>` on all `agy` CLI executions, ensuring that the process working directory (`fake_home` or scenario `work_dir`) is registered as an active workspace.

### DIFF
--- a/evalbench/generators/models/agy_cli.py
+++ b/evalbench/generators/models/agy_cli.py
@@ -506,7 +506,7 @@ class AgyCliGenerator(AgentCliGenerator):
 
         env = self._merged_env()
         cmd = self._base_agy_command(
-            self.agy_bin, "ping", model=self.model
+            self.agy_bin, "ping", model=self.model, add_dirs=[self.fake_home]
         )
         try:
             subprocess.run(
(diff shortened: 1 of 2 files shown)


evalbench/generators/models/agy_cli.py
```diff
@@ -506,7 +506,7 @@ def _verify_runtime(self, configured_servers: list):
 
         env = self._merged_env()
         cmd = self._base_agy_command(
-            self.agy_bin, "ping", model=self.model
+            self.agy_bin, "ping", model=self.model, add_dirs=[self.fake_home]
```
silviojr, line 509: I am not sure if wee need to pass the add_dirs here.

PR #585, opened 2026-08-31
fix(agy): register work_dir as an agy workspace via --add-dir

omkargaikwad23:
## Summary

  `work_dir` was only passed to agy as the subprocess cwd. agy uses the cwd
  for file reads, but runs its shell and write tools in `<appDataDir>/scratch`,
  so files the agent created never landed in `work_dir`. Only `--add-dir` sets
  the tool directory.

  This passes `--add-dir <work_dir>` in addition to the cwd, emitted only when
  a scenario declares `work_dir` — the MCP startup probe and existing scenarios
  are unaffected. The `fake_home` cwd fallback is not mirrored into the flag,
  since it holds harness internals rather than scenario data.

  Only agy needs this. `claude_code`, `codex_cli` and `gemini_cli` honour the
  process cwd directly.

  ## Test plan

  - [x] `pytest evalbench/test/agy_cli_test.py` — 68 passed
  - [x] New test pins the full path: `create_command(cwd=...)` → `--add-dir` → subprocess cwd
  - [x] The no-`work_dir` case is covered by existing exact-argv tests, which fail if the flag is
  always emitted
  - [x] Real eval run with `work_dir` set: the agent's file landed in `work_dir` and scratch was empty
  (before the fix, the reverse)
  - [x] Multi-turn: turn two still runs in `work_dir`

  The rubric scorer only reads the transcript, so the scenario scored 100%
  before and after. The fix was verified on disk.

### DIFF
--- a/evalbench/generators/models/agy_cli.py
+++ b/evalbench/generators/models/agy_cli.py
@@ -833,7 +833,7 @@ class AgyCliGenerator(AgentCliGenerator):
     def _base_agy_command(
         cli: str, prompt: str, resume: bool = False, model: str = None,
         output_format: str = None, log_file: str = None,
-        timeout: str = None,
+        timeout: str = None, add_dir: str = None,
     ) -> list:
         """Builds the non-interactive ``agy -p`` argv shared by the eval
         turn path and the setup-time MCP probe.
@@ -852,6 +852,11 @@ class AgyCliGenerator(AgentCliGenerator):
 
         ``timeout`` maps to ``--print-timeout`` (e.g. "20m"). If omitted,
         defaults to agy's internal default (5 minutes).
+
+        ``add_dir`` maps to ``--add-dir``, registering the scenario's
+        ``work_dir`` as an agy workspace. Required on top of the subprocess
+        cwd: unregistered, agy runs its shell and write tools in
+        ``<appDataDir>/scratch`` and agent writes miss ``work_dir``.
         """
         command = [cli, "-p", prompt, "--dangerously-skip-permissions"]
         if model:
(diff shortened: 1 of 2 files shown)


evalbench/generators/models/agy_cli.py
```diff
@@ -913,8 +920,11 @@ def _run_agy_cli(self, cli_cmd: CLICommand, timeout_seconds=None):
         command = self._base_agy_command(
             self.agy_bin, cli_cmd.prompt, cli_cmd.resume, self.model,
             output_format="stream-json", log_file=self.cli_log_path,
-            timeout=self.timeout,
+            timeout=self.timeout, add_dir=cli_cmd.cwd,
         )
+        # The fallback is deliberately not mirrored into --add-dir: fake_home
+        # holds harness internals (agy binary, settings, MCP config, session
+        # state), which the agent has no business editing as workspace files.
```
prernakakkar-google, line 927:
```suggestion
        # The fallback is deliberately not mirrored into --add-dir: fake_home
        # holds the sandbox ADC copy, agy binary, settings, MCP config and
        # session state. Scenarios that seed env_files into fake_home without a
        # work_dir therefore still write to scratch .
```

prernakakkar-google:
Please correct/rephrase the description.

evalbench/generators/models/agy_cli.py
```diff
@@ -854,6 +854,11 @@ def _base_agy_command(
 
         ``timeout`` maps to ``--print-timeout`` (e.g. "20m"). If omitted,
         defaults to agy's internal default (5 minutes).
+
+        ``add_dir`` maps to ``--add-dir``, registering the scenario's
```
prernakakkar-google, line 858: add in docs as well.

prernakakkar-google (approved):

evalbench/generators/models/agy_cli.py
omkargaikwad23: Thanks for the approval. Docs are already covered ref: https://github.com/GoogleCloudPlatform/evalbench/blame/7d11688c8bfa49e98e62e096a1d7ad73b6157bc0/docs/configs/agentic-dataset-config.md#L42

PR #596, opened 2026-09-10
ci: add harness integration CI for the four agent CLIs

omkargaikwad23:
## Summary

Adds `.ci/` — a Cloud Build pipeline running integration smoke tests for all four agent CLIs (`agy`, `Claude Code`, `Codex`, `Gemini`) against the real Cloud SQL Admin MCP endpoint. Existing coverage relies on unit tests with mocked subprocesses, which cannot catch CLI flag changes, self-updates, or output format shifts.

### Key Design Choices

- **Shared Evalset & Run Config:** Standardizes on a single evalset canonicalized to `<server>__<tool>` via `tool_naming.py`, and a parameterized run config keyed by `CI_HARNESS` so results remain comparable across harnesses.
- **Two Read-Only Scenarios:** A list-then-get query, and a get query against a non-existent instance (zero provisioning and no leaked test infrastructure on failure).
- **Parallel Execution:** Harness build steps execute concurrently with `allowFailure` so failures in one CLI do not cancel the others.
- **Authentication:** `agy` uses a service-account key fetched from Secret Manager to carry `quota_project_id`; the other three CLIs authenticate via build credentials.
- **Unpinned CLI Versions:** The three npm-installed CLIs track `@latest` and `agy` self-updates, so the build surfaces version drift as it ships rather than when a pin is bumped.
- **Scope:** Real MCP servers only (skill/plugin installation smoke tests will follow in a separate PR).

### Verification Gate (`verify_harness.py`)

Because `evalbench.eval()` exits 0 on run completion, `verify_harness.py` enforces a two-tier evaluation gate:

| Tier | Scorers | Rule |
|---|---|---|
| **Non-zero** | 7 telemetry scorers + `trajectory_matcher` | `> 0` |
| **Liveness** | 3 LLM judges | row exists, no error, numeric |

- **Why non-zero for telemetry/trajectory?** Telemetry scorers swallow parse errors and return `0.0` inside `comparison_logs` instead of raising an error. Gating `trajectory_matcher > 0` verifies that expected MCP tools were actually invoked rather than falling back to shell execution.
- **Why liveness only for judges?** Judges are checked for execution health only; non-deterministic judge verdicts are never hard-gated to avoid pipeline flakiness.

---

## Test Plan

- [x] Tested `verify_harness.py` against synthetic results (healthy passes, Tier 1 at 0 fails, judges at 0 pass).
- [x] Verified evalset placeholder expansion (unset variables raise errors; no literal `${...}` reaches agent).
- [x] Verified shared run config resolution across all four harnesses (correct model config, generator, and output paths).
- [x] Triggered real Cloud Build run (`verify-harnesses`) and confirmed execution across all harnesses.

### DIFF
--- a/.ci/harness_smoke.evalset.json
+++ b/.ci/harness_smoke.evalset.json
@@ -1,0 +1,31 @@
+{
+  "scenarios": [
+    {
+      "id": "ci-csql-list-and-get",
+      "starting_prompt": "list all instances in project ${EVAL_GCP_PROJECT_ID}",
+      "conversation_plan": "Ask the agent to list instances in project ${EVAL_GCP_PROJECT_ID}. Once all instances are listed if ${CI_CSQL_INSTANCE} exist get its state and validate its RUNNABLE",
+      "expected_trajectory": [
+        "cloud-sql__list_instances",
+        "cloud-sql__get_instance"
+      ],
+      "env": {
+        "GOOGLE_CLOUD_PROJECT": "${EVAL_GCP_PROJECT_ID}"
+      },
+      "kind": "tools",
+      "max_turns": 3
+    },
+    {
+      "id": "ci-csql-get-instance-not-found",
+      "starting_prompt": "Get the details for the Cloud SQL instance named 'ci-missing-instance-do-not-create' in project '${EVAL_GCP_PROJECT_ID}'.",
+      "conversation_plan": "The user wants details for an instance that does not exist. The agent should call get_instance once, receive a not-found error, and report to the user that the instance could not be found. The user accepts that answer and ends the conversation. The user must not ask the agent to list instances, search for similar names, or otherwise recover.",
+      "expected_trajectory": [
+        "cloud-sql__get_instance"
+      ],
+      "env": {
+        "GOOGLE_CLOUD_PROJECT": "${EVAL_GCP_PROJECT_ID}"
+      },
+      "kind": "tools",
+      "max_turns": 2
+    }
+  ]
+}
--- a/.ci/harness_smoke.evalset.json
+++ b/.ci/harness_smoke.evalset.json
@@ -3,7 +3,7 @@
     {
       "id": "ci-csql-list-and-get",
       "starting_prompt": "list all instances in project ${EVAL_GCP_PROJECT_ID}",
-      "conversation_plan": "Ask the agent to list instances in project ${EVAL_GCP_PROJECT_ID}. Once all instances are listed if ${CI_CSQL_INSTANCE} exist get its state and validate its RUNNABLE",
+      "conversation_plan": "The user wants to see every Cloud SQL instance in project ${EVAL_GCP_PROJECT_ID}, then check the health of one of them. Once the agent has listed the instances, the user asks for the current state of ${CI_CSQL_INSTANCE} and expects the agent to report whether it is RUNNABLE. The user accepts that answer and ends the conversation.",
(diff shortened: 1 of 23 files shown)


.ci/model_configs/claude_code_model.yaml
```diff
@@ -0,0 +1,23 @@
+claude_code_version: "@anthropic-ai/claude-code@2.1.85"
```
prernakakkar-google: check against latest

.ci/model_configs/claude_code_model.yaml
```diff
@@ -0,0 +1,23 @@
...
+generator: claude_code
+model: "claude-opus-4-6"
```
prernakakkar-google: opus 5

.ci/model_configs/gemini_cli_model.yaml
```diff
@@ -0,0 +1,19 @@
+gemini_cli_version: "@google/gemini-cli@0.55.1"
```
prernakakkar-google: latest

.ci/harness_smoke.evalset.json
```diff
@@ -0,0 +1,31 @@
...
+      "conversation_plan": "The user wants to see every Cloud SQL instance in project ${EVAL_GCP_PROJECT_ID}, then check the health of one of them. Once the agent has listed the instances, the user asks for the current state of ${CI_CSQL_INSTANCE} and expects the agent to report whether it is RUNNABLE. The user accepts that answer and ends the conversation.",
+      "expected_trajectory": [
+        "cloud-sql__list_instances",
+        "cloud-sql__get_instance"
+      ],
+      "env": {
+        "GOOGLE_CLOUD_PROJECT": "${EVAL_GCP_PROJECT_ID}"
+      },
+      "kind": "tools",
+      "max_turns": 3
```
prernakakkar-google: 2

prernakakkar-google:
Lets write a design document on this first.
We should use different GCP Project as its a security risk since data is visble publically

.ci/model_configs/claude_code_model.yaml
```diff
@@ -0,0 +1,23 @@
...
+vertex_project_id: !ENV ${EVAL_GCP_PROJECT_ID}
+# Not us-central1: Vertex serves Claude from only a few regions.
+vertex_region: "us-east5"
+
+env:
+  GOOGLE_CLOUD_PROJECT: !ENV ${EVAL_GCP_PROJECT_ID}
+  GOOGLE_CLOUD_LOCATION: "us-central1"
+
+setup:
+  mcp_servers:
```
prernakakkar-google, line 14: we should test for skills as well

omkargaikwad23: That would be covered in followup CL (I have added a note in PR description)

omkargaikwad23:
> Lets write a design document on this first. We should use different GCP Project as its a security risk since data is visble publically

Thanks for reviewing the changes!
External contributors won't be able to view the build logs or access or run the trigger. Same GCP project is used for current test trigger and for crema evals as well. To keep triggers in same gcp project, I decided to use the same project. Will share the doc soon.

.ci/model_configs/claude_code_model.yaml
omkargaikwad23: Testing skills as well for codex and claude code

prernakakkar-google (approved):

PR #601, opened 2026-09-11
fix(claude_code): measure per-tool latency, which was always reported as 0

omkargaikwad23:
## Summary

Issue found during harness CI check work: [logs](https://screenshot-v2.corp.google.com/0pocopjjpqvdg)

`tool_call_latency` scored 0 for every Claude Code scenario. `tools.durationMs` and `totalDurationMs` were initialized to 0 in `_parse_stream_json` and never accumulated — and there was nothing to accumulate from: Claude's stream-json carries no timestamps, and the CLI ran to completion before parsing, so nothing measured the gap.

Fix: stream stdout line-by-line, stamp `time.monotonic()` when a `tool_use` arrives and again at its matching `tool_result`, then accumulate into the tool stats.

Also guards the parser against a malformed line: `_stamp_tool_event` assumed every decoded line was an object, so a line parsing to a list raised `AttributeError`, broke out of the stdout drain loop, and discarded the rest of the agent's response.

## Notes for reviewers

- `_execute_cli_streaming` mirrors `codex_cli._execute_cli_command` verbatim. Hoisting both into `AgentCliGenerator` is the right end state, left as a follow-up to keep this bugfix reviewable. `codex_cli` also has the same `AttributeError` hole.
- `_execute_cli_command` is unchanged — its two setup callers still expect a bare `CompletedProcess`.
- Parallel calls share a start time, so `totalDurationMs` sums tool time rather than wall-clock. That matches the other generators and how the scorer already aggregates.

## Test plan

- [x] `pytest evalbench/test/claude_code_test.py` — 7 passed, up from 5
- [x] Covers the regression itself, both `tool_result` shapes, parallel calls, and malformed lines
- [x] Full suite: no new failures
- [x] Confirm `tool_call_latency > 0` for `claude_code` in the next harness CI run

### DIFF
--- a/evalbench/generators/models/claude_code.py
+++ b/evalbench/generators/models/claude_code.py
@@ -639,13 +639,11 @@ class ClaudeCodeGenerator(AgentCliGenerator):
         self, command: list[str], env: dict[str, str] | None = None,
         cwd: str | None = None, timeout_seconds: float | int | None = None,
     ) -> tuple[subprocess.CompletedProcess, dict[str, int]]:
-        """Runs the Claude Code CLI with line-streamed stdout so we can stamp
-        the wall-clock time at which each stream-json event arrives.
+        """Runs the CLI with line-streamed stdout, returning the usual
+        CompletedProcess plus a `{tool_use_id: duration_ms}` map.
 
-        Returns the usual CompletedProcess plus a `{tool_use_id: duration_ms}`
-        map — measured as the gap between a `tool_use` block and its matching
-        `tool_result`. Claude's stream-json events carry no timestamps, so this
-        in-process stamping is the only path to per-tool latency.
+        Claude's stream-json carries no timestamps, so stamping arrival times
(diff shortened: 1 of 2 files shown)


prernakakkar-google:
WHy verify-harness is showing up

prernakakkar-google:
Parallel tool calls double-count into totalDurationMs

omkargaikwad23:
> WHy verify-harness is showing up

I have added the PR trigger in gcp project. We will be releasing that feature soon.

omkargaikwad23:
> Parallel tool calls double-count into totalDurationMs

Good point! Currently, all four generators and toolcalllatency.py sum parallel calls to measure total tool work rather than wall-clock elapsed time. Keeping it as a sum maintains consistency with the other harnesses, but I'm happy to open a follow-up PR across all four generators if we want wall-clock semantics.

prernakakkar-google (approved):

PR #605, opened 2026-09-15
ci: verify the sdist too, and make the package checks fail loudly

omkargaikwad23:
Follow-up to #597 to strengthen package verification in CI.

### Changes
* **Build and verify sdist:** Both wheel and sdist are now built (`uv build`), verified (`twine check --strict`), and tested by importing all modules in isolated virtual environments (`wheelcheck` and `sdistcheck`). Previously, only the wheel was built and tested.
* **Strict Twine check:** Added `--strict` so any metadata warnings fail the check.
* **Verbose CLI failure:** Printed `google-evalbench --help` output to stdout so entrypoint errors are visible in logs instead of failing silently on `grep`.
* **Generalize `verify_package.py`:** Updated log messages from "wheel" to "package" to reflect testing both distribution types.

### Notes
* The `|| true` in the CLI check is required because `absl-py` intentionally exits with status `1` after printing `--help`.

### Test Plan
- [x] Ran full sequence locally and verified in CI:
  - `uv build` generates both wheel and sdist
  - `twine check --strict` passes on both distributions
  - All modules import cleanly from clean wheel virtualenv
  - All modules import cleanly from clean sdist virtualenv
  - CLI `--help` entrypoint check passes
- [x] Negative test: verified that removing `evalproto/*_pb2.py` in the sdist virtualenv fails with `ImportError`.


prernakakkar-google:
Can we disable the trigger from PR till we rollout the PR this can be confusing for contributors

omkargaikwad23:
> Can we disable the trigger from PR till we rollout the PR this can be confusing for contributors

Sg! I have disabled it now. It should be disappeared from new PR / updated PRs.

.github/workflows/verify-package.yml
```diff
@@ -18,18 +18,42 @@ jobs:
...
         enable-cache: true
         cache-dependency-glob: "uv.lock"
-    - name: Build wheel
-      run: uv build --wheel --out-dir dist
+
+    - name: Build distribution packages (wheel + sdist)
+      run: uv build --out-dir dist
+
+    - name: Validate PyPI package metadata
+      run: uvx twine check --strict dist/*
```
prernakakkar-google, line 32: Should we pin version for twine here?

prernakakkar-google (approved):

.github/workflows/verify-package.yml
omkargaikwad23: Left it unpinned intentionally, twine check validates against PyPI's current rules, so pinning would freeze us to stale ones

PR #607, opened 2026-09-16
fix(k8s): cap HPA maxReplicas at 1 for ReadWriteOnce volume

omkargaikwad23:
Update `evalbench_service/k8s/hpa.yaml` to align autoscaler parameters with volume storage constraints.

### Changes
* Set `maxReplicas: 1` (previously `10`).
* Update `averageUtilization` from `50` to `70`.
* Add explicit `behavior` stabilization windows (`120` seconds for scale-up and `900` seconds for scale-down).

### Technical Rationale
* **Storage constraint (`ReadWriteOnce`)**: `evalbench_service/k8s/pvc.yaml` provisions `evalbench-ssd-pvc` with `ReadWriteOnce` access mode (`premium-rwo`). This mode allows only one node to mount the disk at a time. If the HorizontalPodAutoscaler schedules a second pod onto a different node, Kubernetes fails with a `Multi-Attach error for volume`.
* **Why routine deployments did not trigger this**: Standard updates run `make redeploy` (`kubectl rollout restart`), which restarts the pod and does not run `kubectl apply`. However, `make deploy` applies all manifests in `evalbench_service/k8s/`. This update ensures that `make deploy` preserves the active autoscaler limits.

### Validation
* **Client dry-run**: `kubectl apply --dry-run=client -f evalbench_service/k8s/hpa.yaml` succeeds with valid syntax.
* **Live cluster diff**: `kubectl diff -f evalbench_service/k8s/hpa.yaml` exits with code `0` (zero differences against the active cluster state).


prernakakkar-google:
Lets not make this change.

PR #610, opened 2026-09-17
ci(viewer): gate pull requests on the UI loading and rendering

omkargaikwad23:
Nothing loads the UI before deployment, so a change that breaks a page ships undetected. The package check doesn't cover this — the wheel only ships `evalbench*`, so no viewer module is ever imported.

Adds `.ci/verify_viewer.py` and a path-filtered workflow that runs it. The script imports the entrypoint, asserts the expected pages and modules registered, then renders the page, `on_load`, and each tab in-process.

Two things that look like they'd work but don't, hence the approach:

- **HTTP probes are useless here.** Mesop serves the same shell for every URL. `/`, `/summary`, and `/totally/made/up` all return 200, and so does a page whose body unconditionally raises.
- **Catching exceptions isn't enough.** `render_app_content` wraps its whole body in a `try/except` that only logs, so a broken page never raises. The script fails on logged errors too.

Runs on `ubuntu-latest` in ~45s. No credentials, no network, no results data.

## Test plan

Fault-injected each failure mode and confirmed the gate exits non-zero:

| fault | caught by |
|---|---|
| Charts tab raises | logged error |
| Dataset Quality tab raises | logged error |
| `on_load` raises | exception |
| `dashboard.py` import broken | module assertion |
| entrypoint import broken | import step |
| route renamed `/` → `/renamed` | page assertion |

- [x] Clean run passes with no spurious errors
- [x] Passes in a fresh clone, the way CI sees it
- [x] `pycodestyle` clean


.ci/verify_viewer.py
```diff
@@ -0,0 +1,135 @@
+#!/usr/bin/env python3
```
prernakakkar-google, line 1: We are just checking empty page render at times precompute doesnt happen properly and its not being caught so we should check whether data is loaded as well

.ci/verify_viewer.py
```diff
@@ -0,0 +1,135 @@
...
+        # A fresh request context per attempt, so state does not leak between them.
+        captured.records.clear()
+        try:
+            with app.test_request_context():
+                fn()
+        except Exception as e:
+            failures.append(f"{label} raised {type(e).__name__}: {e}")
+            traceback.print_exc()
+        for record in captured.records:
+            failures.append(f"{label} logged an error: {record.getMessage()}")
```
prernakakkar-google, line 87: record.getMessage() drops exc_info, which loses the traceback on exactly the failure this script exists to catch.

prernakakkar-google:
Can do as part of follow up PR checks on Javascript, Event handlers etc.

.ci/verify_viewer.py
omkargaikwad23: Fixed. Included traceback details in logged error reports.

.ci/verify_viewer.py
omkargaikwad23: Good catch. The gate now precomputes a committed fixture and asserts each tab actually draws its run ids and product names, so a cache that loads nothing fails instead of passing as an empty page.

omkargaikwad23:
> Can do as part of follow up PR checks on Javascript, Event handlers etc.

Sure. will send a followup PR for event handlers CI tests.

.ci/verify_viewer.py
```diff
@@ -0,0 +1,223 @@
...
+        viewer_app.on_load(me.LoadEvent(path=ROOT_PAGE))
+        check_data(label)
+
+    def report():
+        print(f"\n{len(failures)} check(s) failed:")
+        for failure in failures:
+            print(f"  {failure}")
+        return 1
+
+    import precompute_trends
```
prernakakkar-google: nit: at the top

.ci/verify_viewer.py
```diff
@@ -0,0 +1,223 @@
...
+
+    def render_tab(tab, label):
+        state = me.state(viewer_app.State)
+        state.selected_main_tab = tab
+        for field, value in TAB_STATE.get(tab, {}).items():
+            setattr(state, field, value)
+        rt.run_path(ROOT_PAGE)
+        check_data(label)
+
+    def fire_on_load(label):
```
prernakakkar-google, line 175: calls rt.run_path(), then on_load(), then check_data() — so the tree being inspected predates on_load and the check_data call is just a duplicate of the render_page("/") attempt. Mesop's real order is on_load → render; swapping the two lines makes the check mean something.

.ci/verify_viewer.py
```diff
@@ -0,0 +1,223 @@
+#!/usr/bin/env python3
```
prernakakkar-google, line 1: Can add a timeout if tests fail in a given time

.ci/verify_viewer.py
omkargaikwad23: Moved import precompute_trends up to join the other deferred imports in verify() (line 115), right after import mesop as me. It can't go to module top since it needs sys.path.insert(0, VIEWER_DIR) to run first.

.ci/verify_viewer.py
omkargaikwad23: Good catch - swapped, and seeded a job_id query param first (Mesop's real order) so the render actually asserts on_load's state instead of re-checking the plain render.

.ci/verify_viewer.py
omkargaikwad23: Added timeout-minutes: 10 on the job - covers the uv install too, and fails fast instead of default timeout duration

prernakakkar-google (approved):

PR #611, opened 2026-09-17
ci(nl2sql): gate pull requests on a two-tier smoke check

omkargaikwad23:
## Problem

The pre-merge eval in `cloudbuild.yaml` ran a full benchmark, then asserted one thing: `generated_sql` was non-null and non-empty on every row.

Two blind spots followed. Query execution was never checked, so an execution regression passed as long as the model returned a string. Scorer health was never checked, so a scorer that threw was indistinguishable from one that passed, because failures land in `comparison_error` and nothing read that column.

Coverage was also narrower than the runtime suggested: dozens of DQL prompts across multiple trials, which measures model variance rather than exercising more code.

## Change

A curated eight-prompt SQLite set spanning DQL, DML and DDL (`.ci/nl2sql_smoke.evalset.json`), run by `.ci/nl2sql_run_config.yaml` and gated by `.ci/verify_nl2sql.py`.

- **Tier 1, structural.** `returned_sql` and `executable_sql` must score above zero for `dql`, `dml` and `ddl` **separately**. A single aggregate would be carried by the DQL rows while DML and DDL failed outright. Row existence per query type is its own assertion, since `oneshotorchestrator.py` turns a failed sub-dataset into a log line and `eval()` still exits 0.
- **Tier 2, liveness.** Every other scorer named in the run config, judges included, must have a row for each eval, a finite numeric score, and an empty `comparison_error`. The verdict is not gated; a threshold on judge output makes the build flaky on ordinary model variance.

`generated_error` is deliberately not checked. A model writing invalid SQL is a scored outcome, and `executable_sql` already maps it to 0.

The scorer list is read from the run config, so a scorer added there is covered automatically. `num_trials: 2` is load-bearing: trials apply only to DQL, and the consistency comparators do not run below 2.

`verifier/verify.py` is removed.

## Bug this caught

`copy_for_dialect` resolves `setup_sql` and `cleanup_sql` into lists, but the DML branch of `sqlexecwork.py` passed them straight to `db.execute`, whose first act is `query.strip()`. The `AttributeError` escaped before `golden_result` was assigned, so every scorer then failed with `KeyError: 'golden_result'` on every DML row. The DDL branch two lines below already unwrapped these; that is now shared as `_resolve_sql`, which additionally handles a dialect dict whose value is a list.

## Test plan

- [x] Full suite against real Gemini: 12 evals (dql 8, dml 2, ddl 2), verifier exits 0 with 68 checks across 7 scorers.
- [x] Gate fails, exit 1 with a readable message, for each of: DDL `returned_sql` forced to 0; all DDL rows deleted; `comparison_error` set on a judge row; `nan` in a Tier 1 score.
- [x] `evalbench/test/sqlexecwork_test.py` — 3 of its 4 cases fail against the unfixed DML branch.
- [x] `pytest evalbench/test` — 639 passed. The 11 `agent_grpc_proxy` failures also fail on a clean `main`.


prernakakkar-google:
@erinlimbogoogle can you please validate if these smoke tests are relevant one.

PR #613, opened 2026-09-17
feat(mcp_readability): record the tool surface each run was judged against

akangsha7:
**Summary**

A readability result row recorded what the judge concluded but not what it read. That makes a change in finding counts unattributable: an edited tool and a re-run over an identical tool surface produce the same kind of row, and nothing distinguishes them. The judge is a thinking model, so even at temperature 0 an unchanged surface can yield different findings from one run to the next — without a record of the input, there is no way to tell the two apart after the fact.

Rows now carry the endpoint's stable key, a hash per tool, and a hash of the whole tool surface. Nothing reads these columns yet; they are written unconditionally so the history is already in place for whatever does.

**Changes**

1. `generators/models/mcp_tool_formatter.py` — extracts `format_tool_section` from the loop inside `format_tools_to_man_page`, which now returns the join of exactly those sections. No change to the rendered output. Tools are hashed by their rendered section rather than by their `Tool` fields, which makes "the fingerprints match" and "the judge would read identical text" the same statement; hashing fields instead would need a list of which fields matter, kept in sync with this renderer by hand.

2. `scorers/mcp_fingerprint.py` (new) — `tool_fingerprints` maps each tool name to the sha256 of its section, ignoring a later duplicate name to match the first-declaration-wins merge the tools generator already applies across sources. `toolset_fingerprint` hashes the `(name, fingerprint)` pairs sorted by name, so re-declaring a source in a different order does not force a re-judge; the per-tool hashes already carry the ordering guarantee. Imports nothing from concrete scorers, so the scorers and the orchestrator can both use it without a cycle.

3. `evaluator/mcp_readability/orchestrator.py` — adds `endpoint_key`, `toolset_fingerprint`, `tool_fingerprints_json` and `check_timestamp_utc` to `BASE_COLUMNS` and populates them per endpoint. `_endpoint_key` derives identity from the fields the row already reports, so no config change is needed; the cost is that renaming a product orphans its history, and an explicit `id:` on the endpoint survives renames. `check_timestamp_utc` exists because the current timestamp is naive local time, so rows written from different timezones interleave wrongly when sorted.

4. `test/mcp_fingerprint_test.py` (new), `test/mcp_tool_formatter_test.py`, `test/mcp_readability_test.py` — cover the stability properties the hashes are relied on for: a description or schema edit changes a tool's fingerprint, an unrelated tool does not, reordering does not change the toolset hash while adding or renaming does, and each section appears verbatim in the man page. The end-to-end test asserts the columns land on the row and that the toolset hash is derived from the per-tool map rather than computed separately.

**Testing**

`uv run --with pytest pytest evalbench/test -q` and `pycodestyle . --config=.pycodestyle`.


akangsha7:
Recreating from a branch on the main repo so the follow-up baseline PR can stack on it; a fork branch cannot be used as a PR base here.

PR #617, opened 2026-09-18
fix(agy): count skills activated by reading SKILL.md

omkargaikwad23:
## Summary

`extract_skills` looked for an `activate_skill` tool. agy has no such tool — a skill activates when the model reads its `SKILL.md`, and later turns re-use it by running its scripts. So the method always returned `[]` and `skills_trajectory` scored 0.0 for every agy run.

It now matches skill paths in the turn's tool calls, filtered against the skills actually installed in the sandbox.

**Generator**

- **New evalset** `agy-cli-skills.evalset.json`. No shipped agy evalset declared `expected_skills`, which is why this went unnoticed. `example_run_skills_config.yaml` pointed at the non-skills dataset and never enabled `skills_trajectory`; both fixed.
- **Pinned the plugin to `#0.4.0`.** `cloud-sql-postgresql` ships no `skills/` directory on its default branch — they're generated per release — so cloning the bare URL installed a plugin with zero skills.
- **Trimmed the docstrings in `tool_naming.py`** and corrected the agy native-tool examples. Comments only, no behaviour change.

**Scoring**

Turning `skills_trajectory` on for agy exposed three bugs downstream, all fixed here:

- **`skills_best_practices` never found agy's skills.** It searched a fixed list of candidate directories, took the first that existed, and otherwise fell back to hardcoded `.venv/fake_home_*` paths — so a run could be scored against another harness's sandbox. Roots are now globbed per harness layout under the run's own `fake_home`, and agy's plugin layout is covered.
- **A missing `skills_dir` crashed the eval.** `os.listdir` on a configured directory that doesn't exist raised `FileNotFoundError` out of `compare()`. That skill now scores 0 instead.
- **The grade was reported as 0%.** `skills_best_practices` returns a mean out of 100 but was missing from `non_binary_metrics`, so it was matched exactly against 100. A run grading 92.5 and 95.0 printed `0/2 = 0%`.

## Test plan

- [x] `pytest evalbench/test/agy_cli_test.py evalbench/test/skillsbestpractices_test.py` — 76 pass, including new tests that fail on the old code
- [x] `pycodestyle evalbench/ --config=.pycodestyle`
- [x] Live run against Cloud SQL: `skills_trajectory 0/2` before, `2/2 = 100%` after, with `skills_best_practices` reporting its real average
- [x] Local eval run logs: https://paste.googleplex.com/5205274995064832


prernakakkar-google:
Can you share gpaste of example run

datasets/agy-cli-tools/example_run_skills_config.yaml
```diff
@@ -13,11 +13,12 @@ simulated_user_model_config: datasets/model_configs/gemini_2.5_pro_model.yaml
...
-  # Note: trajectory_matcher and parameter_analysis are omitted here because
-  # skill-driven workflows use a different execution surface. The agent activates
-  # high-level skills rather than directly invoking granular tools.
-  # To evaluate skill execution paths, use the `skills_trajectory` scorer instead.
-
+  # trajectory_matcher and parameter_analysis are omitted: the agent activates
+  # high-level skills rather than invoking granular tools directly.
+  skills_trajectory:
+    enforce_order: false
+  skills_best_practices:
```
prernakakkar-google, line 23: skillsbestpractices resolves skills_dir by probing <fake_home>/.codex/skills, .gemini/skills, .claude/skills. agy's skills live at <fake_home>/.gemini/config/plugins/<plugin>/skills/ so it will always be 0

datasets/agy-cli-tools/example_run_skills_config.yaml
```diff
@@ -13,11 +13,12 @@ simulated_user_model_config: datasets/model_configs/gemini_2.5_pro_model.yaml
...
 scorers:
-  # Note: trajectory_matcher and parameter_analysis are omitted here because
-  # skill-driven workflows use a different execution surface. The agent activates
-  # high-level skills rather than directly invoking granular tools.
-  # To evaluate skill execution paths, use the `skills_trajectory` scorer instead.
-
+  # trajectory_matcher and parameter_analysis are omitted: the agent activates
+  # high-level skills rather than invoking granular tools directly.
+  skills_trajectory:
+    enforce_order: false
```
prernakakkar-google: Can you check with enforce_order: true

datasets/agy-cli-tools/agy-cli-skills.evalset.json
```diff
@@ -0,0 +1,38 @@
+{
+  "scenarios": [
+    {
+      "id": "agy-csql-admin-debug-instance",
+      "starting_prompt": "Check on my databases in project ${EVAL_GCP_PROJECT_ID}.",
+      "conversation_plan": "Ask the agent to list all Cloud SQL instances in the project. Once all instances are listed, pick one and ask the agent to get its details. Acceptance criteria: the agent names the instances it found, and for the one picked reports its state verbatim (for example RUNNABLE) rather than describing it in general terms.",
+      "expected_trajectory": [
+        "list_instances",
```
prernakakkar-google: .js here and elsewhere

datasets/agy-cli-tools/example_run_skills_config.yaml
omkargaikwad23: Correct! I have filed an issue - was planning to resolve in separate pr: https://github.com/GoogleCloudPlatform/evalbench/issues/618. Now, I have fixed that issue in same PR.

datasets/agy-cli-tools/example_run_skills_config.yaml
omkargaikwad23: I can see that added in multiple config files - https://github.com/GoogleCloudPlatform/evalbench/blob/5fc047a460f890f05f6419b556bd32f00a57fd37/datasets/codex-skills/codex_skills_run_config.yaml#L13
I will remove it as it's default value is also false.

omkargaikwad23:
> Can you share gpaste of example run

Sure. Attached in the PR description.

prernakakkar-google:
Also validate for other harnesses with cases where skill_dir is mentioned and not mentioned. Validate both on GKE and locally.

datasets/agy-cli-tools/example_run_skills_config.yaml
prernakakkar-google: I want you to validate current changes with enforce_order set to true

omkargaikwad23: Validated the the changes with enforce_order set to true. Logs: Both scenarios score 50, and the analyzer reports skills_trajectory: 0/2 = 0.0% — skills_trajectory isn't in non_binary_metrics, so anything below 100 counts as a fail.
| scenario             | expected | activated          | enforce_order | allow_extra_skills |
|----------------------|----------|--------------------|---------------|--------------------|
| admin-debug-instance | [admin]  | [admin, lifecycle] | 50            | 100                |
| data-schema-explore  | [data]   | [data, admin]      | 50            | 100                |

[Logs Link](https://paste.googleplex.com/5449650438537216)

datasets/model_configs/claude_code_skills_model.yaml
```diff
@@ -10,22 +10,22 @@ model: "claude-opus-4-7"
...
-  # Install skills from main branch (uses npx toolbox-sdk, no local binary needed)
-  # For full plugin marketplace setup (interactive), manually run in Claude Code:
-  #   /plugin marketplace add https://github.com/gemini-cli-extensions/cloud-sql-postgresql.git
-  #   /plugin install cloud-sql-postgresql@cloud-sql-postgresql-marketplace
   skills:
+    # Pinned to a release tag: the default branch ships no `skills/` directory,
+    # they are generated per release.
     - action: install_from_repo
-      url: "https://github.com/gemini-cli-extensions/cloud-sql-postgresql.git"
+      url: "https://github.com/gemini-cli-extensions/cloud-sql-postgresql.git#0.4.0"
```
prernakakkar-google, line 31: Shouldn't we test with latest version?

datasets/codex-skills/codex-skills.evalset.json
```diff
@@ -2,8 +2,8 @@
   "scenarios": [
```
prernakakkar-google, line 2: Can you test skills best practices with codex as well.

prernakakkar-google (approved):

datasets/model_configs/claude_code_skills_model.yaml
omkargaikwad23: #latest isn't valid, the fragment goes straight to git clone --branch, and upstream has no latest ref (tags are
  0.1.7…0.4.0), so it'd fail the clone

prernakakkar-google: When I say latest I mean no tag whatever is latest version

datasets/codex-skills/codex-skills.evalset.json
omkargaikwad23: already tested & attached the logs, please check the line 819 in this log file
https://paste.googleplex.com/4521315336323072#l=819

datasets/model_configs/claude_code_skills_model.yaml
omkargaikwad23: in that case latest version is #0.4.0. 
https://github.com/gemini-cli-extensions/cloud-sql-postgresql/releases

prernakakkar-google: What about when new version is released. the comment means we should not fix.

